PR #7, opened 2016-10-25
Group requirement files.

akatsoulas:
* Add javascript in editorconfig.


requirements/requirements_dev.txt
```diff
@@ -0,0 +1,7 @@
+# Include base requirements
+-r requirements.txt
+
+django>=1.9.6
```
johngian: This is already defined in ``requirements.txt``

PR #10, opened 2016-10-25
Helper method to import settings.

akatsoulas:


mozilla_django_oidc/utils.py
```diff
@@ -0,0 +1,16 @@
...
+
+def import_from_settings(attr):
+    """
+    Load an attribute from the django settings.
+
+    :raises:
+        ImproperlyConfigured
+    """
+    try:
+        setting = getattr(settings, attr)
```
johngian: It would be useful to allow passing a default value to the helper.

PR #23, opened 2016-10-27
Basic authentication functionality.

akatsoulas:


johngian:
s/funcionality/functionality/g

mozilla_django_oidc/auth.py
```diff
@@ -0,0 +1,83 @@
...
+        """Initialize settings."""
+        self.OIDC_OP_TOKEN_ENDPOINT = import_from_settings('OIDC_OP_TOKEN_ENDPOINT')
+        self.OIDC_OP_USER_ENDPOINT = import_from_settings('OIDC_OP_USER_ENDPOINT')
+        self.OIDC_OP_CLIENT_ID = import_from_settings('OIDC_OP_CLIENT_ID')
+        self.OIDC_OP_CLIENT_SECRET = import_from_settings('OIDC_OP_CLIENT_SECRET')
+
+        self.UserModel = get_user_model()
+
+    def verify_token(self, token, **kwargs):
+        """Validate the token."""
```
johngian: What about: ``Verify the token signature.``

mozilla_django_oidc/auth.py
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        # Get the token
+        response = requests.post(self.OIDC_OP_TOKEN_ENDPOINT,
+                                 data=token_payload,
+                                 verify=import_from_settings('VERIFY_SSL', True))
+        # Validate the token
+        payload = self.verify_token(response.get('id_token'))
+
+        user = None
```
johngian: I would prefer the more pythonic paradigm:

```
if foo:
     return x
return None
```

tests/test_auth.py
```diff
@@ -0,0 +1,120 @@
...
+            'verified_email': 'email@example.com'
+        }
+        request_mock.post.return_value = {
+            'id_token': 'id_token',
+            'accesss_token': 'access_token'
+        }
+        self.assertEqual(self.backend.authenticate(code='foo', state='bar'), None)
+
+    def test_get_user(self):
+        """Test get_user method."""
```
johngian: Please add something more descriptive here eg:
``Test get_user method with valid user``

tests/test_auth.py
```diff
@@ -0,0 +1,120 @@
...
+        user = User.objects.create_user('example_username')
+        self.assertEqual(self.backend.get_user(user.pk), user)
+
+    def test_get_invalid_user(self):
+        """Test get_user method."""
```
johngian: Same here

tests/test_auth.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @patch('mozilla_django_oidc.auth.requests')
+    @patch('mozilla_django_oidc.auth.OIDCAuthenticationBackend.verify_token')
+    def test_successful_authentication_existing_user(self, token_mock, request_mock):
+        """Test successful authentication for existing user."""
+        user = User.objects.create_user(username='username',
+                                        email='email@example.com')
+        token_mock.return_value = True
+        request_mock.get.return_value = {
+            'username': 'username',
```
johngian: Please change the username here to be explicitly different from the actual user username.

tests/test_auth.py
```diff
@@ -0,0 +1,120 @@
...
+            'verified_email': 'email@example.com'
+        }
+        request_mock.post.return_value = {
+            'id_token': 'id_token',
+            'access_token': 'access_token'
+        }
+        self.assertEqual(self.backend.authenticate(code='foo', state='bar'), user)
```
johngian: Let's also add some assertion about the mock calls.

tests/test_auth.py
```diff
@@ -0,0 +1,120 @@
...
+        }
+        request_mock.post.return_value = {
+            'id_token': 'id_token',
+            'access_token': 'access_token'
+        }
+        self.assertEqual(User.objects.all().count(), 0)
+        self.backend.authenticate(code='foo', state='bar')
+        self.assertEqual(User.objects.all().count(), 1)
+        user = User.objects.all()[0]
+        self.assertEqual(user.email, 'email@example.com')
```
johngian: Same here.

tests/test_auth.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def test_authenticate_no_code_no_state(self):
+        self.assertEqual(self.backend.authenticate(code='', state=''), None)
+
+    def test_verify_token_no_token(self):
```
johngian: I don't think that this test is really useful for the testcase. You can always hardcode an invalid jwt and test that instead.

tests/test_auth.py
```diff
@@ -0,0 +1,120 @@
...
+        self.assertEqual(self.backend.verify_token(None), None)
+
+    @patch('mozilla_django_oidc.auth.jwt')
+    @patch('mozilla_django_oidc.auth.requests')
+    def test_jwt_decode_params(self, request_mock, jwt_mock):
+        token = jwt.encode({'email': 'email'}, 'secret')
```
johngian: Since we are not actually using this token, this can always be an example string.

mozilla_django_oidc/auth.py
```diff
@@ -0,0 +1,83 @@
+""" This Source Code Form is subject to the terms of the Mozilla Public
```
johngian: Please remove this, we already have a LICENSE file.

mozilla_django_oidc/auth.py
```diff
@@ -0,0 +1,83 @@
...
+        self.OIDC_OP_USER_ENDPOINT = import_from_settings('OIDC_OP_USER_ENDPOINT')
+        self.OIDC_OP_CLIENT_ID = import_from_settings('OIDC_OP_CLIENT_ID')
+        self.OIDC_OP_CLIENT_SECRET = import_from_settings('OIDC_OP_CLIENT_SECRET')
+
+        self.UserModel = get_user_model()
+
+    def verify_token(self, token, **kwargs):
+        """Validate the token."""
+
+        if not token:
```
johngian: We can omit this if, jwt.decode is going to fail anyway.

mozilla_django_oidc/auth.py
```diff
@@ -0,0 +1,83 @@
...
+            return None
+
+        return jwt.decode(token,
+                          self.OIDC_OP_CLIENT_SECRET,
+                          verify=import_from_settings('OIDC_VERIFY_JWT', True))
+
+    def authenticate(self, code=None, state=None):
+        """Authenticates a user."""
```
johngian: Please add a more descriptive docstring eg.
`` Authenticate a user based on OIDC callback code/state ``

mozilla_django_oidc/auth.py
```diff
@@ -0,0 +1,83 @@
...
+    """Override Django's authentication."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialize settings."""
+        self.OIDC_OP_TOKEN_ENDPOINT = import_from_settings('OIDC_OP_TOKEN_ENDPOINT')
+        self.OIDC_OP_USER_ENDPOINT = import_from_settings('OIDC_OP_USER_ENDPOINT')
+        self.OIDC_OP_CLIENT_ID = import_from_settings('OIDC_OP_CLIENT_ID')
+        self.OIDC_OP_CLIENT_SECRET = import_from_settings('OIDC_OP_CLIENT_SECRET')
+
+        self.UserModel = get_user_model()
```
johngian: Instead of defining ``UserModel`` twice please add it in the top of the file.

Issue #31, opened 2016-10-28
Check state string on OIDC callback


johngian:
We should store the random state string generated in OIDC authentication request in user session and verify that the callback has the same string in it to mitigate CSRF, XSS and other attack vectors.

PR #36, opened 2016-10-31
Check against session `oidc_state` on OIDC callback.

johngian:


johngian:
Fixes #31

johngian:
@akatsoulas r?

mozilla_django_oidc/views.py
```diff
@@ -40,6 +41,14 @@ def get(self, request):
                 'code': request.GET['code'],
                 'state': request.GET['state']
             }
+
+            if 'oidc_state' not in request.session:
+                return self.login_failure()
+
+            if request.GET['state'] != request.session['oidc_state']:
+                msg = 'Session `oidc_state` not matching OIDC callback state'
```
akatsoulas: nit: Maybe `Session 'oidc_state' does not match the OIDC callback state`?

mozilla_django_oidc/views.py
```diff
@@ -60,14 +69,18 @@ def __init__(self, *args, **kwargs):
 
     def get(self, request):
         """OIDC client authentication initialization HTTP endpoint"""
+        state = get_random_string(import_from_settings('OIDC_STATE_SIZE', 32))
```
akatsoulas: Although I am in favor of adding this to the request, we should keep that it's recommended and not required to have the `state` in the request.

nit: maybe you could implement this feature as optional, defaulting to True. You can also safely ignore this comment.

johngian: I suggest we keep it this way since its outside of this PRs scope (`state` is already added in the request, we are just checking against the session) and open an issue for that.

PR #38, opened 2016-11-01
Handle user creation.

akatsoulas:
* Use a default username algo.
* Allow users to override the algo through settings.

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -1,5 +1,7 @@
 import base64
+import hashlib
 import jwt
+import logging
 import requests
 
 try:
@@ -7,10 +9,27 @@ try:
 except ImportError:
     from urllib.parse import urlencode
 
+try:
+    from django.utils.encoding import smart_bytes
+except ImportError:
+    from django.utils.encoding import smart_str as smart_bytes
 from django.contrib.auth import get_user_model
 from django.core.urlresolvers import reverse
 
 from mozilla_django_oidc.utils import absolutify, import_from_settings
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def default_username_algo(email):
+    # bluntly stolen from django-browserid
+    # store the username as a base64 encoded sha224 of the email address
+    # this protects against data leakage because usernames are often
+    # treated as public identifiers (so we can't use the email address).
+    return base64.urlsafe_b64encode(
+        hashlib.sha224(smart_bytes(email)).digest()
+    ).rstrip(b'=')
 
 
 class OIDCAuthenticationBackend(object):
@@ -24,6 +43,20 @@ class OIDCAuthenticationBackend(object):
         self.OIDC_OP_CLIENT_SECRET = import_from_settings('OIDC_OP_CLIENT_SECRET')
 
         self.UserModel = get_user_model()
+
+    def create_user(self, email, **kwargs):
+        """Return object for a newly created user account."""
+        # bluntly stolen from django-browserid
+        # https://github.com/mozilla/django-browserid/blob/master/django_browserid/auth.py
+
+        username_algo = import_from_settings('OIDC_USERNAME_ALGO', None)
+
+        if username_algo:
+            username = username_algo(email)
+        else:
+            username = default_username_algo(email)
+
+        return self.UserModel.objects.create_user(username, email)
 
     def verify_token(self, token, **kwargs):
         """Validate the token signature."""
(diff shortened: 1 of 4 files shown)


mozilla_django_oidc/auth.py
```diff
@@ -25,6 +44,32 @@ def __init__(self, *args, **kwargs):
...
+
+        if username_algo:
+            username = username_algo(email)
+        else:
+            username = default_username_algo(email)
+
+        try:
+            return self.UserModel.objects.create_user(username, email)
+        except IntegrityError as err:
+            # Race condition! Attempt to re-fetch from the database.
```
johngian: This is probably more of a ``browserid`` side effect. I suggest we omit this try/except until we have actual cases where django throws an ``IntegrityError``.

mozilla_django_oidc/auth.py
```diff
@@ -25,6 +44,22 @@ def __init__(self, *args, **kwargs):
 
         self.UserModel = get_user_model()
 
+    def create_user(self, email, **kwargs):
+        """Return object for a newly created user account."""
+
+        # bluntly stolen from django-browserid
+        # https://github.com/mozilla/django-browserid/blob/master/django_browserid/auth.py
+        from django.db import IntegrityError  # Importing at the top causes issues on DB init.
```
johngian: Please remove this line.

mozilla_django_oidc/auth.py
```diff
@@ -68,12 +101,28 @@ def authenticate(self, code=None, state=None):
             user_response = requests.get('{url}?{query}'.format(url=self.OIDC_OP_USER_ENDPOINT,
                                                                 query=query))
             user_info = user_response.json()
+            email = user_info.get('email')
+            if email is None:
```
johngian: I would prefer ``if not email`` in case for some reason ``email`` is in dict but is empty.

mozilla_django_oidc/auth.py
```diff
@@ -68,12 +101,28 @@ def authenticate(self, code=None, state=None):
...
+                return self.UserModel.objects.get(email=email)
+            except self.UserModel.MultipleObjectsReturned:
+                # In the rare case that two user accounts have the same email address,
+                # log and bail. Randomly selecting one seems really wrong.
+                LOGGER.warn('Multiple users with email address %s.', email)
+                return None
             except self.UserModel.DoesNotExist:
-                return self.UserModel.objects.create_user(username=user_info['nickname'],
-                                                          email=user_info['email'])
+                pass
```
johngian: Instead of pass why don't you move the user creation related code here?

runtests.py
```diff
@@ -22,6 +22,7 @@
         SITE_ID=1,
         SITE_URL='http://example.com',
         MIDDLEWARE_CLASSES=(),
+        OIDC_USERNAME_ALGO=False
```
johngian: Do we need that? It has a default value anyway.

tests/test_auth.py
```diff
@@ -189,3 +193,92 @@ def test_jwt_decode_params_verify_false(self, request_mock, jwt_mock):
 
         self.backend.authenticate(code='foo', state='bar')
         jwt_mock.decode.assert_has_calls(calls)
+
+    @override_settings(OIDC_CREATE_USER=False)
+    @patch('mozilla_django_oidc.auth.jwt')
+    @patch('mozilla_django_oidc.auth.requests')
+    def test_no_user_creation(self, request_mock, jwt_mock):
```
johngian: Please use something more descriptive like: ``test_user_creation_disabled``

tests/test_auth.py
```diff
@@ -189,3 +193,92 @@ def test_jwt_decode_params_verify_false(self, request_mock, jwt_mock):
...
+        post_json_mock.json.return_value = {
+            'id_token': 'id_token',
+            'access_token': 'access_granted'
+        }
+        request_mock.post.return_value = post_json_mock
+        self.assertEqual(self.backend.authenticate(code='foo', state='bar'), None)
+
+    @patch('mozilla_django_oidc.auth.jwt')
+    @patch('mozilla_django_oidc.auth.requests')
+    def test_user_creation(self, request_mock, jwt_mock):
```
johngian: Same here i would prefer ``def test_user_creation_enabled``.

tests/test_auth.py
```diff
@@ -189,3 +193,92 @@ def test_jwt_decode_params_verify_false(self, request_mock, jwt_mock):
...
+        }
+        request_mock.post.return_value = post_json_mock
+        self.assertEqual(self.backend.authenticate(code='foo', state='bar'),
+                         User.objects.get(email='email@example.com'))
+
+    @patch.object(settings, 'OIDC_USERNAME_ALGO')
+    @patch('mozilla_django_oidc.auth.jwt')
+    @patch('mozilla_django_oidc.auth.requests')
+    def test_custom_username_algo(self, request_mock, jwt_mock, algo_mock):
+        """Test with user creation with custom username algorithm."""
```
johngian: This should probably be: ``Test user creation with custom username algorithm.``

mozilla_django_oidc/auth.py
akatsoulas: I really don't like the idea of having a block of code that does unrelated to the exception stuff inside the `except` clause. If the program flow has reached this point, all we have to do is decide upon the user creation, thus the `pass` here.

runtests.py
akatsoulas: Yes :P

tests/test_auth.py
akatsoulas: That's what the docsting says

runtests.py
johngian: Please use a realistic value that's callable then.

PR #48, opened 2016-11-11
Add a filter_users method in authentication class.

akatsoulas:


mozilla_django_oidc/auth.py
```diff
@@ -105,21 +112,18 @@ def authenticate(self, code=None, state=None):
...
                 # In the rare case that two user accounts have the same email address,
                 # log and bail. Randomly selecting one seems really wrong.
                 LOGGER.warn('Multiple users with email address %s.', email)
                 return None
-            except self.UserModel.DoesNotExist:
-                create_user = import_from_settings('OIDC_CREATE_USER', True)
-
-            if create_user:
+            elif import_from_settings('OIDC_CREATE_USER', True):
                 user = self.create_user(email)
```
johngian: Not completely related but since we are going for a more generic approach I would prefer to allow users to override ``create_user`` and have access to ``claims`` instead of passing just the email.

johngian:
r+wc

mozilla_django_oidc/auth.py
```diff
@@ -105,22 +115,19 @@ def authenticate(self, code=None, state=None):
...
-                return None
 
-            create_user = False
-            try:
-                return self.UserModel.objects.get(email=email)
-            except self.UserModel.MultipleObjectsReturned:
+            # email based filtering
+            users = self.filter_users_by_claims(user_info)
+
+            if len(users) == 1:
```
johngian: If ``users == None`` it raises an error because ``object of type 'NoneType' has no len()``

mozilla_django_oidc/auth.py
```diff
@@ -44,12 +44,22 @@ def __init__(self, *args, **kwargs):
...
+
+    def create_user(self, claims):
         """Return object for a newly created user account."""
         # bluntly stolen from django-browserid
         # https://github.com/mozilla/django-browserid/blob/master/django_browserid/auth.py
 
         username_algo = import_from_settings('OIDC_USERNAME_ALGO', None)
+        email = claims.get('email')
+        if not email:
+            return None
```
johngian: This should return an empty list instead.

mozilla_django_oidc/auth.py
```diff
@@ -44,12 +44,22 @@ def __init__(self, *args, **kwargs):
 
         self.UserModel = get_user_model()
 
-    def create_user(self, email, **kwargs):
+    def filter_users_by_claims(self, claims):
+        """Return all users matching the specified email."""
+        email = claims.get('email')
+        if not email:
+            return None
```
johngian: Same here.

mozilla_django_oidc/auth.py
johngian: Or an empty ``queryset`` if we want to be more consistent.

mozilla_django_oidc/auth.py
```diff
@@ -105,22 +115,19 @@ def authenticate(self, code=None, state=None):
...
                 # log and bail. Randomly selecting one seems really wrong.
                 LOGGER.warn('Multiple users with email address %s.', email)
                 return None
-            except self.UserModel.DoesNotExist:
-                create_user = import_from_settings('OIDC_CREATE_USER', True)
-
-            if create_user:
-                user = self.create_user(email)
+            elif import_from_settings('OIDC_CREATE_USER', True):
+                user = self.create_user(user_info)
```
johngian: Nit: you can always ``return self.create_user(user_info)``, but feel free to omit.

mozilla_django_oidc/auth.py
akatsoulas: This holds true for the `filter_users_by_claims`. The create_user does not return a queryset but and object.

johngian: 👍

mozilla_django_oidc/auth.py
akatsoulas: This should be fixed by the empty queryset.

PR #52, opened 2016-11-16
Add Nonce support.

akatsoulas:


akatsoulas:
@johngian r?

mozilla_django_oidc/auth.py
```diff
@@ -99,7 +110,7 @@ def authenticate(self, code=None, state=None):
 
         # Validate the token
         token_response = response.json()
-        payload = self.verify_token(token_response.get('id_token'))
+        payload = self.verify_token(token_response.get('id_token'), **{'nonce': nonce})
```
johngian: I think its less obscure to got with:

``
payload = self.verify_token(token_response.get('id_token'), nonce=nonce)
``

mozilla_django_oidc/auth.py
```diff
@@ -65,21 +67,30 @@ def create_user(self, claims):
...
 
-    def authenticate(self, code=None, state=None):
+    def authenticate(self, **kwargs):
         """Authenticates a user based on the OIDC code flow."""
 
+        code = kwargs.pop('code', None)
+        state = kwargs.pop('state', None)
+        nonce = kwargs.pop('nonce', None)
+
         if not code or not state:
```
johngian: Shouldn't we also check here ``if nonce`` too?

mozilla_django_oidc/auth.py
```diff
@@ -65,21 +67,30 @@ def create_user(self, claims):
...
             secret = base64.urlsafe_b64decode(self.OIDC_RP_CLIENT_SECRET)
 
-        return jwt.decode(token, secret,
-                          verify=import_from_settings('OIDC_VERIFY_JWT', True),
-                          audience=audience)
+        id_token = jwt.decode(token, secret,
+                              verify=import_from_settings('OIDC_VERIFY_JWT', True),
+                              audience=audience)
+
+        if import_from_settings('OIDC_USE_NONCE') and nonce != id_token['nonce']:
```
johngian: Please add a default value here like you do in the rest of reference of ``OIDC_USE_NONCE``.

mozilla_django_oidc/auth.py
akatsoulas: Nonce is optional so we should proceed with token validation regardless

mozilla_django_oidc/auth.py
akatsoulas: Completely agree. The only reason for going for keyword args, is because I didn't want to add the nonce keyword in the function signature. It's optional so if anybody wants to use it can pass as a keyword arg. Thoughts?

mozilla_django_oidc/auth.py
akatsoulas: If it is true and somehow `nonce=None` then the token validation will fail.

PR #54, opened 2016-11-18
Improve token validation.

akatsoulas:


johngian:
Code looks OK. I've also tested it locally and JWT verification works fine. r+

PR #59, opened 2016-11-22
Add a method to refresh the id_token.

akatsoulas:


akatsoulas:
@johngian r?

mozilla_django_oidc/views.py
```diff
@@ -109,6 +109,7 @@ class OIDCLogoutView(View):
 
     @property
     def redirect_url(self):
+        """Return the logout url defines in settings."""
```
johngian: ``s/defines/defined/g``

johngian:
It would be great if you can add a test to check the output given an ``id_token``.

PR #74, opened 2017-01-04
Pass the request object in the authenticate method.

akatsoulas:


mozilla_django_oidc/auth.py
```diff
@@ -86,10 +86,14 @@ def verify_token(self, token, **kwargs):
     def authenticate(self, **kwargs):
         """Authenticates a user based on the OIDC code flow."""
 
-        code = kwargs.pop('code', None)
-        state = kwargs.pop('state', None)
+        request = kwargs.pop('request', None)
+        if not request:
+            return None
```
johngian: I suggest we return ``HttpResponseBadRequest``

mozilla_django_oidc/auth.py
```diff
@@ -86,10 +86,14 @@ def verify_token(self, token, **kwargs):
...
+
+        state = request.GET.get('state')
+        code = request.GET.get('code')
         nonce = kwargs.pop('nonce', None)
-        session = kwargs.pop('session', None)
+        session = request.session
 
         if not code or not state:
             return None
```
johngian: Same here.

tests/test_auth.py
```diff
@@ -139,13 +148,21 @@ def test_successful_authentication_new_user(self, token_mock, request_mock, algo
     def test_authenticate_no_code_no_state(self):
         """Test authenticate with wrong parameters."""
 
-        self.assertEqual(self.backend.authenticate(code='', state=''), None)
+        # there are no GET params
+        request = RequestFactory().get('/foo')
+        request.session = {}
+        with self.assertRaises(SuspiciousOperation) as context:
```
johngian: Instead of checking the message in a different assertion you can always use:
self.assertRaisesMessage(exception, message)

PR #86, opened 2017-04-28
Add support for Django 1.11 and Python 3.6

### LINKED ISSUE #85: add support for django 1.11
Django 1.11 (LTS) released in April 2017.

This issue covers adding support for it.

That might be as easy as adding relevant bits in the `tox.ini`.

willkg:
* add test environments covering Django 1.11 and Python 3.6
* fix tox environments so they test correctly
* remove requirements.txt which is an out-of-sync copy of what's in setup.py

Fixes #85.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@
 language: python
 
 python:
+  - "3.6"
   - "3.5"
   - "3.4"
   - "3.3"
@@ -22,6 +23,12 @@ matrix:
       env: DJANGO="Django<1.11"
     - python: "3.3"
       env: DJANGO="Django<1.12"
+    - python: "3.6"
+      env: DJANGO="Django<1.9"
+    - python: "3.6"
+      env: DJANGO="Django<1.10"
+    - python: "3.6"
+      env: DJANGO="Django<1.11"
 
 before_install:
   - pip install codecov
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -1,3 +1,0 @@
-django>=1.9.6
-requests
-python-jose>=1.3.2
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,11 @@ setup(
         'mozilla_django_oidc',
     ],
     include_package_data=True,
-    install_requires=['Django>1.7', 'python-jose', 'requests'],
+    install_requires=[
+        'Django>1.7',
+        'python-jose',
+        'requests'
+    ],
     license='MPL 2.0',
     zip_safe=False,
     keywords='mozilla-django-oidc',
@@ -54,6 +58,7 @@ setup(
         'Framework :: Django :: 1.8',
         'Framework :: Django :: 1.9',
         'Framework :: Django :: 1.10',
+        'Framework :: Django :: 1.11',
         'License :: OSI Approved :: Mozilla Public License 2.0 (MPL 2.0)',
         'Intended Audience :: Developers',
         'Natural Language :: English',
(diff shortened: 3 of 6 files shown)


.travis.yml
```diff
@@ -22,6 +23,12 @@ matrix:
       env: DJANGO="Django<1.11"
     - python: "3.3"
       env: DJANGO="Django<1.12"
+    - python: "3.6"
+      env: DJANGO="Django<1.9"
+    - python: "3.6"
+      env: DJANGO="Django<1.10"
+    - python: "3.6"
+      env: DJANGO="Django<1.11"
```
willkg: Pretty sure I got this right.

Django 1.11 is the first version to support Python 3.6--none of the other Django versions do. So I added exclusions for those combinations.

requirements/requirements.txt
```diff
@@ -1,3 +0,0 @@
-django>=1.9.6
-requests
-python-jose>=1.3.2
```
willkg: This is redundant with what's in `setup.py`. Seems better to just `pip install -e .` rather than maintain two copies of things. This reduces the likelihood that they become out of sync again.

setup.py
```diff
@@ -44,7 +44,11 @@
         'mozilla_django_oidc',
     ],
     include_package_data=True,
-    install_requires=['Django>1.7', 'python-jose', 'requests'],
+    install_requires=[
+        'Django>1.7',
```
willkg: I suspect this should be `Django>=1.8`. Does this library support Django 1.7? If so, there's no testing for it.

tox.ini
```diff
@@ -3,13 +3,15 @@ envlist =
...
     PYTHONPATH = {toxinidir}:{toxinidir}/mozilla_django_oidc
 commands = python runtests.py
 deps =
-    django-18: Django>=1.8,<1.9
-    django-19: Django>=1.9,<1.10
-    django-110: Django>=1.10
+    django18: Django>=1.8,<1.9
+    django19: Django>=1.9,<1.10
+    django110: Django>=1.10,<1.11
+    django111: Django>=1.11
```
willkg: These were specified wrong originally because these bits don't have a hyphen. I think tox was ignoring that and just keeping whatever the `requirements.txt` file installed. Thus tox was just testing the same Django version over and over again.

setup.py
johngian: You are right, we don't support ``1.7``. I guess this happened because we assumed that ``>1.7`` means ``>=1.8`` which is wrong.

johngian:
We have an auth0 specific middleware under ``contrib``:

* https://github.com/mozilla/mozilla-django-oidc/blob/master/mozilla_django_oidc/contrib/auth0/middleware.py

This should be adapted in order to be compatible with Django 1.10 and 1.11 middleware changes. 

That said it didn't work with Django 1.10 anyway so i am not sure if this fix is relevant to this PR. Also, I am not sure if we use it anymore or if it's something we should continue maintaining under this project.

* https://docs.djangoproject.com/en/1.11/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware

@akatsoulas what do you think?

.travis.yml
johngian: 👍

requirements/requirements.txt
johngian: 👍

willkg:
If we spin the middleware issue (not supported in django 1.10+ and probably isn't tested) off to a new issue, I can look into that next.

johngian:
Sounds good to me.

PR #87, opened 2017-04-28
Allow combining OIDCAuthenticationBackend with other backends.

johngian:
* Return `None` if request not available in kwargs

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -88,7 +88,7 @@ class OIDCAuthenticationBackend(object):
 
         self.request = kwargs.pop('request', None)
         if not self.request:
-            raise SuspiciousOperation('Request object not found.')
+            return None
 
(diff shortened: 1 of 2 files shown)


johngian:
@willkg r?

willkg:
Pretend I don't know much about Django auth. When would `request` not be in the args to `authenticate`?

akatsoulas:
Please ignore the approval. I realized that @willkg has already started this after the review submission.

willkg:
@akatsoulas I don't know much about that aspect of the Django authentication API, so if you're good with this, that's ok with me. I can look into it later today, but haven't, yet.

johngian:
The idea is that different backends pass different args as credentials (eg. ModelBackend passes user/pass, OIDC passes a request object). In order to be able to fallback to different backends you need to be able to check if given the args, authentication returns a User. If not fallback to next available backend.

Issue #91, opened 2017-05-01
User username field shouldn't get a bytes value


willkg:
The Django username field should get a unicode value in Python 2 and a string value in Python 3.

However, it's getting the output of `default_username_algo` which generates a string in Python 2 (which is fine since there are no non-ascii characters in it) and a bytes in Python 3 (which is probably not fine).

Further, the signature for username generation functions is "weird" in the sense that it takes a string/bytes as an argument and returns a string/bytes--but it operates on an email address and returns a username. That feels weird.

This issue covers fixing this part of the API.

willkg:
I claim we should do the following:

1. change the signature for the username generation algorithm so that it takes a string/unicode (it's an email address, so if it's anything else, that's weird) and returns a string/unicode
2. fix the `default_username_algo` accordingly

Does that make sense? If so, I can throw together a fix today.

I can't tell if this is related to #89 since there's no traceback, so I have no idea where that error is occurring.

akatsoulas:
@willkg it totally does make sense!

PR #92, opened 2017-05-01
Add documentation for connecting oidc identities to django users

### LINKED ISSUE #88: document how auth0 identities match to django identities
There's no documentation covering how mozilla-django-oidc takes an auth0 identity from a user_token and figures out which Django user that belongs to.

It's pretty important for any site that has to migrate from an existing system to this one. Further, surfacing how that works will likely surface inherent restrictions regarding accounts.

willkg:
Fixes #88.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -76,3 +76,48 @@ Please add the following to your ``settings.py``:
 
 Finally let your OpenID connect OP know about your callback URL. In our example this is:
 ``http://127.0.0.1:8000/oidc/callback/``.
+
+
+Additional optional configuration
+=================================
+
+Connecting OIDC user identities to Django users
+-----------------------------------------------
+
+By default, mozilla-django-oidc looks up a Django user matching the email field
+to the email address returned in the user info data from the OIDC provider.
+
+This means that no two users in the Django user table can have the same email
+address. Since the email field is not unique, it's possible that this can
+happen. Especially if you allow users to change their email address. If it ever
+happens, then the users in question won't be able to authenticate.
+
+If you want different behavior, subclass the
+:py:class:`mozilla_django_oidc.auth.OIDCAuthenticationBackend` class and
+override the `filter_users_by_claims` method.
+
+For example, let's say we store the email address in a ``Profile`` table
+in a field that's marked unique so multiple users can't have the same
+email address. Then we could do this:
+
+.. code-block:: python
+
+   from mozilla_django_oidc.auth import OIDCAuthenticationBackend
+
+   class MyOIDCAB(OIDCAuthenticationBackend):
+       def filter_users_by_claims(self, claim):
+           email = claims.get('email')
+           if not email:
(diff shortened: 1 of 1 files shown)


willkg:
Does this look ok? What can I/we improve?

docs/installation.rst
```diff
@@ -76,3 +76,48 @@ Please add the following to your ``settings.py``:
...
+For example, let's say we store the email address in a ``Profile`` table
+in a field that's marked unique so multiple users can't have the same
+email address. Then we could do this:
+
+.. code-block:: python
+
+   from mozilla_django_oidc.auth import OIDCAuthenticationBackend
+
+   class MyOIDCAB(OIDCAuthenticationBackend):
+       def filter_users_by_claims(self, claim):
```
willkg: Part of me was wondering whether we should talk about what's in the claim and/or tell the user/developer how to figure out what's in the claim. I'm pretty sure the information in the claim is different depending on whether you're using Mozilla's SSO or Auth0 or some other OIDC provider, so I'm not sure how to do this effectively. Maybe have a paragraph after this along the lines of:

> The information available in the claim depends on which OIDC provider you're using. Best way to figure out what's in the claim is to log the information in a development environment.

Maybe we should also add?:

> The data in the claim isn't static and can change over time and also be different depending on the user that logged in.

Are both of those true in your experience? Should we add them to reduce the surprises users/developers might encounter when trying to integrate mozilla-django-oidc into their sites?

akatsoulas: I believe both these paragraphs will help users/developers to better understand what to expect regarding the data in the claim. 👍 for adding those.

docs/installation.rst
```diff
@@ -76,3 +76,48 @@ Please add the following to your ``settings.py``:
...
+Connecting OIDC user identities to Django users
+-----------------------------------------------
+
+By default, mozilla-django-oidc looks up a Django user matching the email field
+to the email address returned in the user info data from the OIDC provider.
+
+This means that no two users in the Django user table can have the same email
+address. Since the email field is not unique, it's possible that this can
+happen. Especially if you allow users to change their email address. If it ever
+happens, then the users in question won't be able to authenticate.
```
akatsoulas: Does it make sens to add here a couple of lines about the `default_username_algo` and how it is used to populate the django username?

willkg: I think it does, too. I was writing a section about Django user creation, but the API looked weird and so I wrote up #91. I figure when I fix that issue, I'll add some documentation about it, too.

akatsoulas:
@willkg is this ready for merge?

willkg:
@akatsoulas Yes, please!

I'm trying to keep PRs small and topical to make them easier to work through. I can make bigger PRs if that's easier. Just let me know.

akatsoulas:
No need for bigger PRs. I was just wondering for the current one because of the two paragraphs in the code block in your comment. Thanks!

PR #95, opened 2017-05-03
Clean up username generation and document it

### LINKED ISSUE #91: User username field shouldn't get a bytes value

willkg:
* fix default_username_algo to return the correct types and add tests
* add docs regarding how mozilla-django-oidc generates Django User accounts and
  how to change how usernames are generated

Fixes #91.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -171,7 +171,8 @@ is created, then you should subclass the
 :py:class:`mozilla_django_oidc.auth.OIDCAuthenticationBackend` class and
 override the `create_user` method.
 
-For example, let's say you want to create a ``Profile``, too:
+For example, let's say you want to populate the ``User`` instance with other
+data from the claims:
 
 .. code-block:: python
 
@@ -181,7 +182,9 @@ For example, let's say you want to create a ``Profile``, too:
    class MyOIDCAB(OIDCAuthenticationBackend):
        def create_user(self, claims):
            user = super(OIDCAuthenticationRequestView, self).create_user(claims)
-           profile = Profile(user=user, is_new=True)
+
+           user.first_name = claim.get('given_name', '')
+           user.last_name = claim.get('family_name', '')
 
            return user
 
@@ -189,3 +192,8 @@ For example, let's say you want to create a ``Profile``, too:
 Then you'd use the Python dotted path to that class in the
 ``settings.AUTHENTICATION_BACKENDS`` instead of
 ``mozilla_django_oidc.auth.OIDCAuthenticationBackend``.
+
+
+.. seealso::
+
+   https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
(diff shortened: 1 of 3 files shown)


docs/installation.rst
```diff
@@ -121,3 +121,71 @@ email address. Then we could do this:
...
+
+   from mozilla_django_oidc.auth import OIDCAuthenticationBackend
+   from myapp.models import Profile
+
+   class MyOIDCAB(OIDCAuthenticationBackend):
+       def create_user(self, claims):
+           user = super(OIDCAuthenticationRequestView, self).create_user(claims)
+           profile = Profile(user=user, is_new=True)
+
+           return user
```
willkg: I don't know if this example is too contrived to be helpful. Is there a better example/use-case we could use here?

johngian: I guess introspecting the OIDC authentication scope and getting the profile data might be more relevant, although this one is also useful.

https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

docs/installation.rst
```diff
@@ -121,3 +121,71 @@ email address. Then we could do this:
 Then you'd use the Python dotted path to that class in the
 ``settings.AUTHENTICATION_BACKENDS`` instead of
 ``mozilla_django_oidc.auth.OIDCAuthenticationBackend``.
+
+
+Creating a Django ``User`` record for new users
+-----------------------------------------------
+
+If a user logs into your site and doesn't already have an account, by default,
```
akatsoulas: nit: Shall we also mention something about the `OIDC_CREATE_USER`?  Eg that the default behavior can be changed by the said variable or does it feel too verbose?

willkg: It's definitely worth mentioning `OIDC_CREATE_USER`, but I don't really understand it.

If you have it set to False, how does the flow work? Don't you end up with a user that's authenticated, but has no corresponding Django user? What do you do at that point?

docs/installation.rst
willkg: Ahhh--yeah! That's a much better use-case. I'll change the example.

docs/installation.rst
akatsoulas: That's a good point. A possible scenario could be the one where the `authenticate` method can be overridden for the use case where a developer needs just to verify an email/user data like eg in a campaign and save them (completely ignoring the Django user). That said, probably this is a hack that does not conform with the Django flow and needs some more work/thought. An idea could be to handle cases like the above at an earlier stage of the flow. @willkg @johngian thoughts?

willkg: I like that there's an "escape hatch", but I think any documentation for the setting should probably include some guidance on why you'd want to use it and how to use it.

I think I want to push off thinking about this and any resolution to another issue/PR and land this one as is in regards to `OIDC_CREATE_USER`.

Is that ok? If so, I'll write up an issue about it.

willkg: @akatsoulas I wrote up #96 to cover documenting `OIDC_CREATE_USER`.

willkg:
I think we've worked through all the outstanding issues in this PR.

Did I miss anything?

PR #101, opened 2017-05-09
Re-organize installation to clarify things

willkg:
* add section on getting a client id and secret
* note that you need a client id and secret for every environment you're using
* break up settings changes so it's clearer what comes from where
* add warning about client id and secret being secret


willkg:
Travis failed, but it has nothing to do with the changes since they're all docs changes. The problem is from installing requests 2.14:

https://github.com/kennethreitz/requests/issues/4006

PR #103, opened 2017-05-10
Decode base64 secret to string instead of bytes.

johngian:


willkg:
Are there any tests that cover that code? Seems like we should have one because that kind of thing is different between Python 2 and Python 3.

If there aren't tests, then we should either add one in this PR or write up an issue to add one.

johngian:
I am thinking that it doesn't make a lot of sense of manually handling the encoding/decoding to base64 programmatically. If an OIDC provider uses base64 decoding we can have a doc entry on how to prepare the value by decoding the secret in CLI. This way we remove some logic + code from the lib.

@akatsoulas @willkg thoughts?

willkg:
I'm not sure I know enough about this and what it entails to know whether it's something that's important to have in the library or not.

Is it something the Mozilla SSO is doing?

johngian:
cc @gdestuynder 

Do you think that it makes sense to handle base64 encoded secret keys in the lib's codebase or just add some docs on how to easily encode/decode (or even just omit everything).

Some context around this:
* When we started integrating with the auth0 service, client secret where base64 encoded.
* At some point this stopped happening for new apps connected to auth0.
* We added this config entry to easily enable/disable base64 support per app.

gdestuynder:
Hi,
we discussed this issue during our tech alignment meeting

basically - the base64 code here is to work-around an auth0 specific issue, that auth0 has fixed. clients with a secret that is not url safe needs their secret to be regenerated though (I believe it is only mozillians.org right now), which can be done by request (we can either regenerate it and send the new one, or make a new client if that makes the transition either).

I would recommend this path forward (getting a new secret that is) and getting rid of the code, as it does not seem to be part of OIDC's spec http://openid.net/developers/specs/

johngian:
Closing this one and will open another PR that removes everything related to base64.

Issue #90, opened 2017-04-28
Make auth0 middleware compatible with Django>1.10


johngian:
Starting from version 1.10, Django introduced a new style for middleware classes.
https://docs.djangoproject.com/en/1.11/releases/1.10/#new-style-middleware

At the moment ``contrib.auth0.middleware`` is not compatible nor tested. We should make sure to adapt it to these changes.

willkg:
I'll work on this today.

PR #104, opened 2017-05-11
Add RefreshIDToken middleware documentation

willkg:


willkg:
Seems to make sense to put it in the "optional" section, but maybe it doesn't?

I think I got the language right. If it's unclear, we should fix it.

Also, is that the only thing someone needs to do to add the middleware to their site? Are there any other ordering requirements in the `MIDDLEWARE_CLASSES`?

Also also, I'm pushing off the Django 1.11-related bits to be covered in issue #90.

akatsoulas:
Regarding the language it was pretty clear to me what this does. 

I am struggling to decide whether this should be optional or if we need to create a section for Auth0 specific configuration. On the other hand, because this raises awareness about persisting sessions in a site, I believe we should keep it as is. @johngian any thoughts on the matter?

akatsoulas:
The middleware order looks good to me. I am merging this so that we proceed to other issues and if there is anything missing or wrong, we can amend the documentation.

PR #105, opened 2017-05-17
tox everything, fixes #100

### LINKED ISSUE #100: tox everything
The README suggests running tests with python.
Apparently the travis file uses its own matrix.
Ideally everything should be all tox. 

Also, the documentation should guide developers how to run individual tests rapidly without having to run the whole suite across all versions. 

I'm happy to work on this.

peterbe:
This PR tidies up everything around testing. 

* `tox` to rule them all. Meaning, that's what travis runs and that's the first time developers are asked to run
* `tox` now runs the linting too. Both of `tests/` and of `mozilla_django_oidc/`
* The `.travis.yml` is simplified and uses `tox-travis` which makes sure all envs are run in parallel
* Uses `django-admin.py test` to run tests instead of `runtests.py`
* The settings is now its own file in `tests/settings.py`


peterbe:
Hmm... apparently that's now how you start the `[testenv:lint]` from travis. I need to figure that out.

peterbe:
Note, only the Python env is parallelized. Each python version runs multiple versions of Django.

tox.ini
```diff
@@ -1,17 +1,41 @@
...
+    py{27,34,35}-django{19,110}
+    py{27,34,35,36}-django111
+
+[travis]
+python =
+  2.7: py27
+  3.3: py33
+  3.4: py34
+  3.5: py35
+  3.6: py36, coverage, lint
```
peterbe: This is the key trick! When you run travis it runs the environments one by one. But we only bother with coverage and linting if you're using python 3.6.

peterbe:
Can someone help me understand why the test coverage dropped? Or is it not true?

johngian:
Can you try removing ``--branch`` from coverage command? It's the only difference that looks like it might be related to coverage report. Also its not true since we are not changing any code :)

willkg:
Can we move this one forward? It'll make testing middleware changes for Django 1.11 (issue #90) easier.

@peterbe is at a conference today so he's not around. I'm happy to do a follow-up PR to fix is outstanding including the `--branch` argument.

peterbe:
Removing the `--branch` did fix it. Locally I got 93% with `--branch` and 96% without. 

Merge?

PR #111, opened 2017-05-23
90 auth0 middleware

willkg:


mozilla_django_oidc/urls.py
```diff
@@ -16,3 +18,16 @@
...
+    # If we're running tests, it helps to have a "home page" we can test
+    # against that doesn't trigger authentication things.
+    from django.http import HttpResponse
+
+    def fakeview(req):
+        return HttpResponse('Win!')
+
+    urlpatterns.append(
+        url(r'^mdo_fake_view/$', fakeview, name='mdo_fake_view')
+    )
```
willkg: I don't know a better way of doing this where "better" doesn't involve creating a django project in the tests or something equally large and finicky across Django versions.

I'm pretty sure the `if 'test' in sys.argv` will work, but if there's a better way to determine we're in a test run, I'm all ears.

tests/auth0_tests/test_middleware.py
```diff
@@ -61,17 +80,119 @@ def test_stale_cache_valid_token(self, mock_cache, mock_refresh):
...
+            raise Exception('Unable to authenticate with %r' % credentials)
+
+        ret = super(ClientWithUser, self).login(**credentials)
+        if not ret:
+            raise Exception('Login failed')
+
+        # Stash the user object it used and rebuild the UserifiedClientHandler
+        self.user = user
+        self.handler = UserifiedClientHandler(self.enforce_csrf_checks, user=self.user)
+        return ret
```
willkg: This was pretty rough. Here's what's going on.

Django's test client doesn't set `request.user` before merrily jaunting off into yonder field of daisies and Django's `RequestFactory` doesn't go through middleware at all.

Since we really want a few tests that test handling a request with the mechanics of the middleware to verify that the Django 1.10+ transitional mixin worked with our middleware correctly, so I was forced to do this weirdo thing where we set our own `ClientHandler` that adds a `.user` to the request before passing it through the machinery.

Another way to do this might be to switch to webtest, but I was too emotionally drained to see what shenanigans I would have to pull from my magic hat to get that vaguely working.

So that's why it looks so wonky. I tried to add lots of comments to alleviate the weirdness. Let me know if it's still unclear.

Also, I *hate* how Django's test client `.login` fails silently. That's so irritating and I lost at least an hour tinkering with code that was fine trying to debug it. In what bizarre and paranormal circumstances would I ever want it to fail silently, I cannot begin to fathom.

willkg:
I think the coveralls report shows a decrease because the branch occurs with different versions of Django and I think it's only looking at one tox environment run. If that theory is right, then I don't think there's anything we can do to make it better and I'm not inclined to spend time trying.

willkg:
Also, if there's anything else from #90 that's not covered here, let me know.

mozilla_django_oidc/urls.py
johngian: What if we subclass ``TestCase`` and add an extended urlconf same way you do it here. That means:

* Import and extend urls under /test
* Configure TestCase to use the patched urlconf

https://docs.djangoproject.com/en/1.11/topics/testing/tools/#urlconf-configuration

If it feels to much of an overhead let's leave it as it is.

willkg: Oo--I didn't know about that. I'll try it out.

willkg:
@johngian ^^^ That implements your idea. Thank you! That's a much better way to do it.

johngian:
Please squash the last 2 commits and i can merge it.

willkg:
^^^ Squashed.

PR #115, opened 2017-05-30
Fix docs errors

willkg:
* fix the client id and secret settings
* fix the jinja2 template example


willkg:
I'm in the process of testing things out on Standup and hit these two mistakes in the docs.

Issue #94, opened 2017-05-02
Support prompt parameter none instead of the delegation endpoint


gdestuynder:
prompt=none is the OIDC standard and is now supported by auth0:
https://auth0.com/docs/api-auth/tutorials/silent-authentication

Sample (Lua):
https://github.com/pingidentity/lua-resty-openidc/commit/2d21cb521f5edebb76af27bfb1f7a44264d8372e

In other words, instead of calling the delegation endpoint (auth0 specific) we pass a prompt=none parameter to the authorization endpoint, which will return the current id_token (just like delegation endpoint) without prompting the user

gdestuynder:
Same for Flask-pyoidc: https://github.com/zamzterz/Flask-pyoidc/pull/17

willkg:
I was going to work on this today. Pretty sure it's straight forward. The link to the flask equivalent helps.

peterbe:
@gdestuynder Pardon my ignorance but back in the day when we used `/delegation` instead to renew the ID token we were able to do that **from the server**. I.e. a communication between, say, airmozilla and auth0. The client would not even notice, at all, when things go well. In @willkg 's implementation in https://github.com/mozilla/mozilla-django-oidc/issues/127 he makes the **user** redirect to renew the ID token if it needs to be renewed. If all goes well, the redirect should just flash by and redirect back to where you were and all the user sees is a delay and flashing pages that zip by. 

Is there not a way to do the renewal of the ID token without forcing the user to use her browser to go there?

If the answer, to the above question, is no; can you help us how with a solution to this: https://github.com/mozilla/mozilla-django-oidc/pull/127#issuecomment-305525863

Perhaps the answer is to implement auth0 in JavaScript and do the ["silent authenticatio within a hidden iframe"](https://github.com/auth0/auth0.js#api).

willkg:
@peterbe https://github.com/mozilla/mozilla-django-oidc/blob/master/mozilla_django_oidc/contrib/auth0/middleware.py#L29

It was excluding AJAX-y things already.

peterbe:
@willkg that's because the code originated from airmozilla which was/is a very server-side rendering heavy site. 

[...]

PR #120, opened 2017-05-31
Fix middleware docs

willkg:
The docs were missing two additional things you had to have in your settings.
This adds them.


willkg:
I realize this is moot if we work on #94, but no one has gotten to that, yet.

willkg:
I need a working system, but then I think I can tackle #94.

PR #122, opened 2017-05-31
Fix create_user example

willkg:
This fixes a bug I created when writing up the example. Now it should work.


willkg:
Actually, I'm going to bundle this with some other docs fixes. Closing this out for the moment.

PR #123, opened 2017-05-31
always redirect to logout_url even when logged out, fixes #121

### LINKED ISSUE #121: UnboundLocalError on logout
See traceback below.
Happens when I try to go to `http://localhost:8000/oidc/logout/`

The [relevant code](https://github.com/mozilla/mozilla-django-oidc/blob/320792bb6b2c95d4257ff34c9897b67ddf6ced29/mozilla_django_oidc/views.py#L118-L128) assumes that the user is authenticated. 

```
web_1          | Traceback (most recent call last):
web_1          |   File "/usr/local/lib/python3.5/wsgiref/handlers.py", line 137, in run
web_1          |     self.result = application(self.environ, self.start_response)
web_1          |   File "/usr/local/lib/python3.5/site-packages/django/contrib/staticfiles/handlers.py", line 63, in __call__
web_1          |     return self.application(environ, start_response)
web_1          |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/wsgi.py", line 157, in __call__
web_1          |     response = self.get_response(request)
web_1          |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 124, in get_response
web_1          |     response = self._middleware_chain(request)
web_1          |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/exception.py", line 43, in inner
web_1          |     response = response_for_exception(request, exc)
web_1          |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/exception.py", line 93, in response_for_exception
web_1          |     response = handle_uncaught_exception(request, get_resolver(get_urlconf()), sys.exc_info())
web_1          |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/exception.py", line 41, in inner
web_1          |     response = get_response(request)
web_1          |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
web_1          |     response = self._get_response(request)
web_1          |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 187, in _get_response
web_1          |     response = self.process_exception_by_middleware(e, request)
web_1          |   File "/usr/local/lib/python3.5/site-packages/django/core/handlers/base.py", line 185, in _get_response
web_1          |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
web_1          |   File "/usr/local/lib/python3.5/site-packages/django/views/generic/base.py", line 68, in view
web_1          |     return self.dispatch(request, *args, **kwargs)
web_1          |   File "/usr/local/lib/python3.5/site-packages/mozilla_django_oidc/views.py", line 128, in dispatch
web_1          |     return HttpResponseRedirect(logout_url)
web_1          | UnboundLocalError: local variable 'logout_url' referenced before assignment
```

peterbe:
CC @willkg

### DIFF
--- a/mozilla_django_oidc/views.py
+++ b/mozilla_django_oidc/views.py
@@ -114,15 +114,15 @@ class OIDCLogoutView(View):
 
     def dispatch(self, request, *args, **kwargs):
         """Log out the user."""
-
-        if request.user.is_authenticated():
+        # Check if a method exists to build the URL to log out the user
+        # from the OP.
+        logout_from_op = import_from_settings('OIDC_OP_LOGOUT_URL_METHOD', '')
+        if logout_from_op:
+            logout_url = import_string(logout_from_op)()
+        else:
             logout_url = self.redirect_url
 
-            # Check if a method exists to build the url to logout the user from the OP
-            logout_from_op = import_from_settings('OIDC_OP_LOGOUT_URL_METHOD', '')
-            if logout_from_op:
-                logout_url = import_string(logout_from_op)()
-
-            # Log out the Django user.
+        if request.user.is_authenticated():
+            # Log out the Django user, only if she was actually logged in.
             auth.logout(request)
         return HttpResponseRedirect(logout_url)
--- a/mozilla_django_oidc/views.py
+++ b/mozilla_django_oidc/views.py
@@ -114,15 +114,16 @@ class OIDCLogoutView(View):
 
     def dispatch(self, request, *args, **kwargs):
         """Log out the user."""
-        # Check if a method exists to build the URL to log out the user
-        # from the OP.
-        logout_from_op = import_from_settings('OIDC_OP_LOGOUT_URL_METHOD', '')
-        if logout_from_op:
-            logout_url = import_string(logout_from_op)()
-        else:
-            logout_url = self.redirect_url
+        logout_url = self.redirect_url
 
         if request.user.is_authenticated():
+            # Check if a method exists to build the URL to log out the user
+            # from the OP.
+            logout_from_op = import_from_settings('OIDC_OP_LOGOUT_URL_METHOD', '')
+            if logout_from_op:
+                logout_url = import_string(logout_from_op)()
+
             # Log out the Django user, only if she was actually logged in.
             auth.logout(request)
+
         return HttpResponseRedirect(logout_url)
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -7,6 +7,7 @@ from mock import patch
 
 from django.core.exceptions import SuspiciousOperation
 from django.contrib.auth import get_user_model
+from django.contrib.auth.models import AnonymousUser
 from django.core.urlresolvers import reverse
 from django.test import RequestFactory, TestCase, override_settings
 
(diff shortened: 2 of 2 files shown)


mozilla_django_oidc/views.py
```diff
@@ -114,15 +114,15 @@ def redirect_url(self):
 
     def dispatch(self, request, *args, **kwargs):
         """Log out the user."""
-
-        if request.user.is_authenticated():
+        # Check if a method exists to build the URL to log out the user
+        # from the OP.
+        logout_from_op = import_from_settings('OIDC_OP_LOGOUT_URL_METHOD', '')
```
akatsoulas: I am not sure whether we should log out the user from the OP's session in the case where the user in not logged in in the app. Probably, the expected behavior would be to do that only when the user is logged in and the relevant setting is present. @peterbe @johngian @willkg thoughts?

willkg: I agree--if the user isn't logged in to the app, then this view should essentially be a no-op.

peterbe: Oh you're right! It shouldn't. That means that only `logout_url = self.redirect_url` can be used if the user is already anonymous. I'll update.

peterbe: Actually. Why not? Maybe. 
What's a good example of an implementation of `OIDC_OP_LOGOUT_URL_METHOD`? Does it assume that the user *is* authenticated. 

Another option altogether is to do something like this:

```python
    def dispatch(self, request, *args, **kwargs):
        """Log out the user."""
        if not request.user.is_authenticated():
            return HttpResponseBadRequest('already signed out')
        logout_from_op = import_from_settings('OIDC_OP_LOGOUT_URL_METHOD', '')
        if logout_from_op:
            logout_url = import_string(logout_from_op)()
        else:
            logout_url = self.redirect_url
        auth.logout(request)
        return HttpResponseRedirect(logout_url)           
```

What do you think about that solution?

peterbe: Ultimately, what I want to avoid is getting 500 Internal Server Errors based on something the client can do. E.g. `curl http://example.com/oidc/logout/`. So we need something to deal with this view being executed by anonymous users.

peterbe: > I agree--if the user isn't logged in to the app, then this view should essentially be a no-op.

But it's not that simple. If the user isn't logged in, sure, you don't call `auth.logut()` but you still need to redirect somewhere. If `OIDC_OP_LOGOUT_URL_METHOD` assumes that the user is authenticated, then it'd be weird to bypass that and fallback on the default (`self.redirect_url`). 

Options:

1. Change the *documentation* such that... "If you implement and set 'OIDC_OP_LOGOUT_URL_METHOD' you can not assume that the user is logged in."
2. My other suggestion above about exiting early with a `HttpResponseBadRequest`. 
3. Return a `403 Forbidden` if someone tries to log out when they're not logged out. 
4. Refactor the patch so that `OIDC_OP_LOGOUT_URL_METHOD` is only invoked when the user is logged in.

willkg: The issue is that if the user is not logged into the app, we don't want to log the user out of the *OP*. Best to just redirect to the `self.redirect_url` in that case.

akatsoulas: I believe the simplest solution would be option 4. Logout the user from the OP only if the setting is present and the user is logged in and redirect to the `self.redirect_url` in all other cases.

Another option could be:
```python
if not user.is_authenticated() and import_from_settings('OIDC_OP_LOGOUT_URL_METHOD', ''):
    return HttpResponseBadRequest('already signed out')
```
This way we won't bypass the setting and at the same time we can silently redirect to `self.redirect_url` when the setting is missing.

peterbe: I'm ok with that. Will update the PR after the meeting.

peterbe: Done. I don't think it's glorious but it's a start and at least things will not raise UnboundLocalError if you sign out as anonymous.

akatsoulas:
@peterbe can you squash those commits into one?

peterbe:
@akatsoulas I just used the GitHub "Squash and Merge" green button.

PR #127, opened 2017-06-01
#94 rewrite RefreshIDToken middleware to use prompt=none

willkg:
This reworks the `RefreshIDToken` middleware to use oidc authorize endpoint with prompt=none.

I also moved all of it from contrib/auth0/ into the root of the project since it's no longer auth0 specific.

I updated docs and tests.


willkg:
I'm pretty sure I got this right. There's some duplication between the view and the middleware, but it's minor and it's unlikely we'll use it in a third place.

I tested it with Standup. One thing I noted is that it's impossible to tell if it's working right. I think I want to add `logging.debug` calls in a bunch of places so we can more easily trace it when verifying things are working.

Does that sound ok? We don't seem to do that anywhere else in the library, but I think we should.

mozilla_django_oidc/middleware.py
```diff
@@ -0,0 +1,82 @@
...
+
+# FIXME(willkg): This doesn't appear to be used anywhere.
+def logout_url():
+    """Log out the user from Auth0."""
+    url = 'https://' + import_from_settings('OIDC_OP_DOMAIN') + '/v2/logout'
+    url += '?' + urlencode({
+        'returnTo': import_from_settings('LOGOUT_REDIRECT_URL', '/'),
+        'client_id': import_from_settings('OIDC_RP_CLIENT_ID')
+    })
+    return url
```
willkg: This doesn't appear to be used anywhere so I added a FIXME to it. If it turns out it is used, then I can remove it.

mozilla_django_oidc/middleware.py
```diff
@@ -0,0 +1,82 @@
...
+class RefreshIDToken(MiddlewareMixin):
+    """Renews id_tokens after expiry seconds
+
+    For users authenticated with an id_token, we need to check that it's still
+    valid after a specific amount of time and if not, force them to
+    re-authenticate silently.
+
+    """
+    def process_request(self, request):
+        if ((request.method == 'GET' and
```
willkg: Because the middleware now redirects to the authorization endpoint if the token has expired, pretty sure we need to restrict it to just GET requests.

Does that sound right? Are there other methods that will work with the redirects?

mozilla_django_oidc/middleware.py
```diff
@@ -0,0 +1,82 @@
...
+             request.user.is_authenticated() and
+             not request.path.startswith('/oidc/') and
```
willkg: I added this because we don't want it to kick off during OIDC related requests like `/oidc/callback`.

mozilla_django_oidc/middleware.py
```diff
@@ -0,0 +1,82 @@
...
+             not request.is_ajax())):
+            expiration = request.session.get('oidc_id_token_expiration')
+            if expiration is not None and expiration > time.time():
```
willkg: `time.time()` will give back seconds since epoch for that server. If all the servers in a cluster are in the same timezone, that'll work fine. If they're not then this won't work.

What's a better way to do this? Should we be using `YYYY-mm-dd HH:MM:SS` UTC timestamps and string comparisons?

mozilla_django_oidc/middleware.py
```diff
@@ -0,0 +1,82 @@
...
+                    'nonce': nonce
+                })
+                request.session['oidc_nonce'] = nonce
+
+            request.session['oidc_state'] = state
+            request.session['oidc_login_next'] = request.get_full_path()
+
+            query = urlencode(params)
+            redirect_url = '{url}?{query}'.format(url=auth_url, query=query)
+            return HttpResponseRedirect(redirect_url)
```
willkg: This method is the meat of the change. If the id token has expired, then this figures out the authentication endpoint url to redirect to and sets up the session and other bits accordingly. Then it redirects and the SSO will redirect back to the callback url if all goes well.

mozilla_django_oidc/middleware.py
peterbe: I don't get this. The epoch is defined with a timezone. It was UTC. 
If I generate a `time.time()` on my server in New York, I get the same number as a server in London. Even though they have different values in `/etc/timezone`.

peterbe: Is it worth saving some characters by doing:
```python
expiration = request.session.get('oidc_id_token_expiration', 0)
if expiration > time.time():
```
??

peterbe: I wonder if this line should be put first in this function. If the expiration is set and it's "future" then there's no point waking up `request.user` or `request.is_ajax` etc.

mozilla_django_oidc/middleware.py
```diff
@@ -0,0 +1,82 @@
...
+            auth_url = import_from_settings('OIDC_OP_AUTHORIZATION_ENDPOINT')
+            client_id = import_from_settings('OIDC_RP_CLIENT_ID')
+            state = get_random_string(import_from_settings('OIDC_STATE_SIZE', 32))
+
+            # Build the parameters as if we were doing a real auth handoff, except
+            # we also include prompt=none.
+            params = {
+                'response_type': 'code',
+                'client_id': client_id,
+                'redirect_uri': absolutify(reverse('oidc_authentication_callback')),
```
peterbe: Perhaps there's something I don't "get" about this `absolutify`. It depends on a config setting which you have to maintain for dev, stage and prod. To accomplish the same result you can just use the request. E.g.

```python
    'redirect_uri': request.build_absolute_uri(reverse('oidc_authentication_callback')),
```

mozilla_django_oidc/middleware.py
```diff
@@ -0,0 +1,82 @@
...
+
+            if import_from_settings('OIDC_USE_NONCE', True):
+                nonce = get_random_string(import_from_settings('OIDC_NONCE_SIZE', 32))
+                params.update({
+                    'nonce': nonce
+                })
+                request.session['oidc_nonce'] = nonce
+
+            request.session['oidc_state'] = state
+            request.session['oidc_login_next'] = request.get_full_path()
```
peterbe: Neat!

mozilla_django_oidc/utils.py
```diff
@@ -2,7 +2,6 @@
     from urlparse import urljoin
 except ImportError:
     from urllib.parse import urljoin
-
 from django.conf import settings
```
peterbe: That break was there to separate standard lib from third-party.

mozilla_django_oidc/middleware.py
```diff
@@ -0,0 +1,82 @@
...
+from django.core.urlresolvers import reverse
+from django.http import HttpResponseRedirect
+from django.utils.crypto import get_random_string
+
+from mozilla_django_oidc.utils import absolutify, import_from_settings
+
+
+# Django 1.11 makes changes to how middleware work. In Django 1.11+, we want to
+# use the mixin so that our middleware works as is.
+if tuple(django.VERSION[0:2]) >= (1, 10):
```
peterbe: First of all, `django.VERSION[0:2]` is already a tuple. And you don't need the slice. 
This is the exact same meaning: `if django.VERSION >= (1, 10):` and it's neater. 

Secondly, the *comment* above says Django 1.11 but this if statement is about Django 1.10.

peterbe: I suspect this is unrelated to the PR at hand. You can ignore this comment and we can deal with it separately.

mozilla_django_oidc/middleware.py
willkg: Ok--sounds like this is fine, then.

mozilla_django_oidc/middleware.py
willkg: I'm mimicking the code in the authentication view.

mozilla_django_oidc/utils.py
willkg: Oops--that shouldn't have been in the PR.

willkg:
> The more I look at this the more I wonder how this will ever work with a modern web app where the Django server just spits out RESTy JSON to a single-page app.

I implemented what @gdestuynder suggested in #94 as best as I could figure. If I didn't do it right, let me know. I have no idea how this works with AJAXy web-apps, but as far as I can see, the existing middleware doesn't work with those, either.

mozilla_django_oidc/middleware.py
```diff
@@ -0,0 +1,92 @@
...
+        pass
+
+
+OIDC_URLS = [
+    reverse(name)
+    for name in [
+        'oidc_authentication_init',
+        'oidc_authentication_callback',
+        'oidc_logout'
+    ]
```
johngian: nit: I think that a config entry to allow devs to extend this list with custom URLs might be handy

mozilla_django_oidc/middleware.py
johngian: Does it make sense to add a catch-all case were we force users to log out ?

mozilla_django_oidc/middleware.py
```diff
@@ -0,0 +1,92 @@
...
+                return
+
+            # The id_token has expired, so we have to re-authenticate silently.
+            auth_url = import_from_settings('OIDC_OP_AUTHORIZATION_ENDPOINT')
+            client_id = import_from_settings('OIDC_RP_CLIENT_ID')
+            state = get_random_string(import_from_settings('OIDC_STATE_SIZE', 32))
+
+            # Build the parameters as if we were doing a real auth handoff, except
+            # we also include prompt=none.
+            params = {
```
johngian: nit: Since we are copying the code from the authentication request, should we write a helper function and call it accordingly?

mozilla_django_oidc/middleware.py
willkg: How would that work?

Would we have it as a setting and have it default to these three? That seems hard to use for developers.

Would we have a setting and always append these three?

mozilla_django_oidc/middleware.py
willkg: I mentioned that in the first PR comment. Namely, there is some duplication, but it'll only be in two places and they have subtle differences, so I'm not sure it helps a lot to centralize it into a helper function.

mozilla_django_oidc/middleware.py
willkg: This forces a re-authentication in a bunch of cases which will result in the user getting logged out. I covered all the cases I could think of, but if there are others, we should make sure they're covered, too.

mozilla_django_oidc/middleware.py
johngian: I was thinking the latter. Having a config entry that defaults to an empty list and we always append these 3 entries.

willkg:
^^^ That rebases against master tip to pick up @peterbe's `is_authenticated` changes. It also fixes nits @peterbe had.

I want to add logging to this because otherwise it's impossible to see if it's working.

I also want someone who knows OIDC really well and what @gdestuynder wanted in #94 to make sure the changes I made are right.

I'll work on adding debug-level logging now.

mozilla_django_oidc/middleware.py
willkg: That makes sense--I'll add that.

mozilla_django_oidc/middleware.py
peterbe: Once this lands, I'll bring it up again. I might need a refresher as to why or why not you can't use the domain that is in the `request` object ...from a security point of view.

willkg:
^ That redoes the exempt urls code and also makes the middleware more flexible in regards to what constitutes a refreshable url.

I also added some light logging which makes it much easier to verify that the middleware is doing something.

I also added some troubleshooting docs.

@johngian Does this look ok? Is the implementation for the `prompt=none` part correct?

mozilla_django_oidc/middleware.py
```diff
@@ -0,0 +1,127 @@
...
+class RefreshIDToken(MiddlewareMixin):
+    """Renews id_tokens after expiry seconds
+
+    For users authenticated with an id_token, we need to check that it's still
+    valid after a specific amount of time and if not, force them to
+    re-authenticate silently.
+
+    """
+    def get_exempt_urls(self):
+        """Generate a return a set of url paths to exempt from RefreshIDToken
```
johngian: typo: Generate and return

johngian:
Code looks OK. Please squash and merge 🚀

willkg:
^^^ That fixes the typo. After that passes, I'll squash and merge.

Thank you!

gdestuynder:
looks alright to me (went through the code quickly)

Issue #102, opened 2017-05-09
match OIDC identity to Django user issues


willkg:
Right now, mozilla-django-oidc matches an OIDC identity to a Django user via the Django user email field.

That's problematic for a couple of reasons:

1. the code is doing a case-sensitive exact match, so it's possible that could fail if the case ever changes (email addresses on some systems are case-insensitive)
2. the email field in the Django user table is not marked as unique and thus you can have multiple Django users with the same email field

I think to fix item 1, we want to do a case-insensitive match even though technically, email addresses are case-sensitive.

I think to fix item 2, we want to be using the username field of the Django user table and not the email field. Django 1.11's username field works fine with that. It means we don't have to generate usernames from email addresses. It means two users can't have the same email address. It solves a bunch of things for us.

Are there other issues we need to work around?

Are there better ways to solve this?

techdragon:
I'd like to recommend that any solution to this issue, provide a simple method for people using custom user models to override the behaviour meant for use with the default Django user model.

For instance one of the things I frequently do with a custom user model is mark the email address as unique in the database and case normalise to lowercase in order to ensure no uppercase/lowercase duplicates.

willkg:
@techdragon It already allows for that.

https://mozilla-django-oidc.readthedocs.io/en/latest/installation.html#connecting-oidc-user-identities-to-django-users

I wrote up this issue to fix the default behavior.

akatsoulas:
Having the email as the identifier is handy because of its use to login to a RP. That said, I think it makes sense to use Django's `username` and maybe we could give the ability to the users to change this behavior through the configuration of the lib.  If we are going down this way, I don't think that we should consider the case-sensitivity of the email an issue any more. 👍 for this change.

willkg:
I'll throw together a PR to fix the case-sensitive check.

I'm on the fence about whether to move this to the username field because Django 1.8 and 1.9 have a username field has a 30 character maximum.

Part of me is thinking that we should wait to make that change until after we've dropped support for Django 1.8 and 1.9.

tpazderka:
Would it be possible to create a mapping model connecting `sub` claim with `UserModel`?

Matching users by `sub` should be more safe and reliable than matching by `username` or `email`.

willkg:
@tpazderka Can't you do that already? https://mozilla-django-oidc.readthedocs.io/en/latest/installation.html#connecting-oidc-user-identities-to-django-users

tpazderka:
[...]

PR #131, opened 2017-06-01
Change email matching to be case-insensitive

willkg:
Emails aren't case sensitive, so we need to match the email address in the claim
with a Django user using a case-insensitive match.

This is part of issue #102.


willkg:
@akatsoulas I don't have permissions to merge--can you or someone else do it?

willkg:
Thank you!

PR #135, opened 2017-06-06
Remove legacy code related to base64 encoded secret. Fixes #89

### LINKED ISSUE #89: Base64 encoding raises error when using Python3
When ``OIDC_RP_CLIENT_SECRET_ENCODED==False`` authentication raises the following error on callback:

```
Exception Type: TypeError at /oidc/callback/
Exception Value: a bytes-like object is required, not 'str'
```

This gets triggered on python 3.x only.

johngian:


johngian:
@akatsoulas r?

mozilla_django_oidc/auth.py
```diff
@@ -79,10 +79,8 @@ def create_user(self, claims):
     def verify_token(self, token, **kwargs):
         """Validate the token signature."""
         nonce = kwargs.get('nonce')
-
         secret = self.OIDC_RP_CLIENT_SECRET
-        if import_from_settings('OIDC_RP_CLIENT_SECRET_ENCODED', False):
-            secret = base64.urlsafe_b64decode(self.OIDC_RP_CLIENT_SECRET)
+
         # Verify the token
         verified_token = jws.verify(token, secret, algorithms=['HS256'])
```
akatsoulas: nit: You can directly use here `self.OIDC_RP_CLIENT_SECRET` since `secret` it's not used anywhere else now that you are removing the base64 stuff so that we can remove another line.

PR #137, opened 2017-06-08
use request instead of SITE_URL, fixes #136

### LINKED ISSUE #136: Use the request instead of SITE_URL
As far as I can see, the only places where we use `SITE_URL` we have access to a `request` object. They have the ability to make absolute URLs without a custom setting. 

I'll make an attempt and refactoring it away but perhaps stop me if there's a need to be able to know the site's public-facing site URL that I don't get.

peterbe:
If we don't *really* need to be able to create absolute URLs in "offline" mode, then I think it's much more convenient to use the `request` to generate absolute URLs in runtime.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -96,13 +96,6 @@ documentation for the appropriate values.
    OIDC_OP_USER_ENDPOINT = "<URL of the OIDC OP userinfo endpoint>"
 
 
-This value depends on your site.
-
-.. code-block:: python
-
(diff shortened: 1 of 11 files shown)


akatsoulas:
@peterbe I will do a review asap unless someone else beats me to it

peterbe:
This would actually help me in my current project because that way I don't have to tell CloudOps to enter another environment variable in the dev, stage and prod deployments. 

I guess what I'm saying is that it would be neat to have a release after this lands :)

PR #139, opened 2017-06-09
[refactoring] smart_bytes exists in django 1.8

peterbe:
It occurred to me that `django.utils.encoding.smart_bytes` works all the way back in Django 1.8.

Also, I changed the `ImportError` tricks to always try Python 3 first.

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -4,8 +4,7 @@ import json
 import logging
 import requests
 
-from django.utils.encoding import smart_bytes
-from django.utils.encoding import smart_text
+from django.utils.encoding import smart_bytes, smart_text
 from django.contrib.auth import get_user_model
 from django.core.exceptions import SuspiciousOperation
 from django.core.urlresolvers import reverse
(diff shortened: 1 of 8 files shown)


peterbe:
oh no. I must have mixed up my branches. Lemme see if I can fix this.

mozilla_django_oidc/auth.py
```diff
@@ -4,10 +4,7 @@
 import logging
 import requests
 
-try:
-    from django.utils.encoding import smart_bytes
-except ImportError:
-    from django.utils.encoding import smart_str as smart_bytes
+from django.utils.encoding import smart_bytes
```
johngian: One line please ``smart_bytes, smart_text``.

setup.py
```diff
@@ -6,6 +6,7 @@
 try:
     from setuptools import setup
 except ImportError:
+    # FIXME(peterbe): In what circumstances do you NOT have setuptools?
```
johngian: This is part of the bootstrapping code generated by ``cookiecutter``. Many projects use it for compatibility because distutils is native python. That said i don't really mind removing this.

johngian:
@peterbe I added some commits. Please squash before merge.
r+wc

PR #144, opened 2017-06-09
only logout with POST, fixes #126

### LINKED ISSUE #126: Only allow logout on POST
We allow logging out simply by GET'ing a URL (`/oidc/logout/`). https://github.com/mozilla/mozilla-django-oidc/blob/04fcbdb1a8697792a7183b99baa0e11beaaa6430/mozilla_django_oidc/views.py#L115

We should probably avoid that. 
E.g. https://stackoverflow.com/a/14587231/205832

peterbe:


johngian:
Please also add a test with a request factory sending a ``post`` to the logout view and check that logout is called as expected.

peterbe:
> Please also add a test with a request factory sending a post to the logout view and check that logout is called as expected.

[This is already one](https://github.com/peterbe/mozilla-django-oidc/blob/3259782487f84669baee2d04540619bb82de9ba5/tests/test_views.py#L322-L328).

tests/test_views.py
```diff
@@ -301,22 +301,17 @@ def test_get(self):
...
         logout_view = views.OIDCLogoutView.as_view()
 
-        with patch('mozilla_django_oidc.views.auth.logout') as mock_logout:
-            response = logout_view(request)
-            mock_logout.assert_called_once_with(request)
-
-        self.assertEqual(response.status_code, 302)
-        self.assertEqual(response.url, '/example-logout')
+        response = logout_view(request)
+        self.assertEqual(response.status_code, 405)
```
peterbe: You're right. It's simply part of Django. No point testing Django. Will remove.

mozilla_django_oidc/views.py
```diff
@@ -123,7 +123,7 @@ def redirect_url(self):
         """Return the logout url defined in settings."""
         return import_from_settings('LOGOUT_REDIRECT_URL', '/')
 
-    def dispatch(self, request, *args, **kwargs):
+    def post(self, request):
```
peterbe: Tests test that automatically. As long as the view doesn't have a `csrf_exempt` decorator it should be using CSRF.

peterbe:
I believe I've address the CSRF question.

PR #152, opened 2017-06-15
Improve security notations in HISTORY

willkg:
* adds security notations to HISTORY which we should continue going forward
* updates release process by adding how to handle security issues and indicators
  of severity

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,20 +5,16 @@ History
 0.3.1 (2017-06-15)
 ++++++++++++++++++
 
-Backwards-incompatible changes:
+Security issues:
 
-* None
-
-Features:
-
-* None
-
-Bugs:
-
-* Sanitize `next` url for the authentication view. Prevents open redirects.
+* **Medium**: Sanitize next url for authentication view
 
 0.3.0 (2017-06-13)
 ++++++++++++++++++
+
+Security issues:
+
+* **Low**: Logout using POST not GET (#126)
 
 Backwards-incompatible changes:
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -89,16 +89,43 @@ Releasing a new version
...
+    Users should upgrade immediately.
+
+  * Medium: vulnerability endangers users by sending them to malicious sites or stealing browser
+    data.
+
+    Users should upgrade immediately.
+
+  * Low: vulnerability is a nuissance to site staff and/or users
+
+    Users should upgrade.
```
willkg: I kind of made this up based on looking at CVE severity calculator (https://www.first.org/cvss/calculator/3.0) and some other things.

Mostly, I think we minimally need a "LOW" which indicates security issues that are more of a nuissance and "HIGH" which indicates security issues that require users to upgrade asap.

I don't want to make this a huge complex thing, but I figured it was helpful to provide guidance to release folks.

What do you think? Are these levels and descriptions helpful? Do they line up with other projects/systems?

HISTORY.rst
```diff
@@ -5,38 +5,29 @@ History
...
-Bugs:
+Security issues:
 
-* Sanitize `next` url for the authentication view. Prevents open redirects.
+* **Medium**: Sanitize next url for authentication view
 
 0.3.0 (2017-06-13)
 ++++++++++++++++++
 
+Security issues:
```
willkg: I'm hoping pyup and requires.io and other tools can key off of this and alert their users appropriately. That's ultimately what we want.

If there are better ways to denote security issues in the HISTORY, I'm all ears.

PR #155, opened 2017-06-22
Forbid access instead of raising error on HTTPError.

### LINKED ISSUE #154: /oidc/callback/ throws exception on 403 from /oauth/token/
```
HTTPError: 403 Client Error: Forbidden for url: https://auth.mozilla.auth0.com/oauth/token
  File "django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 249, in _legacy_get_response
    response = self._get_response(request)
  File "django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "newrelic/hooks/framework_django.py", line 527, in wrapper
    return wrapped(*args, **kwargs)
  File "django/views/generic/base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "newrelic/hooks/framework_django.py", line 914, in wrapper
    return wrapped(*args, **kwargs)
  File "django/views/generic/base.py", line 88, in dispatch
    return handler(request, *args, **kwargs)
  File "mozilla_django_oidc/views.py", line 75, in get
    self.user = auth.authenticate(**kwargs)
  File "django/contrib/auth/__init__.py", line 70, in authenticate
    user = _authenticate_with_backend(backend, backend_path, request, credentials)
  File "django/contrib/auth/__init__.py", line 115, in _authenticate_with_backend
    return backend.authenticate(*args, **credentials)
  File "mozilla_django_oidc/auth.py", line 129, in authenticate
    response.raise_for_status()
  File "requests/models.py", line 937, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
HTTPError: 403 Client Error: Forbidden for url: https://auth.mozilla.auth0.com/oauth/token
```

In our specific case, I'm pretty sure what's going on is that users are using Standups for the first time since I switched from our auth0 library to mozilla-django-oidc and we also recycled the client secret. So the user has a Django session with bits from the previous auth0 setup and it doesn't jive with this one.

However, the 403 is unhandled here and results in Standups returning an HTTP 500 to the user. Seems like mozilla-django-oidc shouldn't do that. Instead, I think it should fail the login and return `None`.

johngian:
Fixes #154

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -126,7 +126,9 @@ class OIDCAuthenticationBackend(object):
         response = requests.post(self.OIDC_OP_TOKEN_ENDPOINT,
                                  data=token_payload,
                                  verify=import_from_settings('OIDC_VERIFY_SSL', True))
-        response.raise_for_status()
+
+        if not response.status_code == requests.codes.ok:
+            return None
 
         # Validate the token
         token_response = response.json()
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -107,6 +107,8 @@ class OIDCAuthenticationBackendTestCase(TestCase):
             'id_token': 'id_token',
             'access_token': 'access_granted'
         }
+        post_json_mock.status_code = 200
+        request_mock.codes.ok = 200
         request_mock.post.return_value = post_json_mock
 
         post_data = {
@@ -148,6 +150,8 @@ class OIDCAuthenticationBackendTestCase(TestCase):
             'id_token': 'id_token',
             'access_token': 'access_granted'
         }
+        post_json_mock.status_code = 200
+        request_mock.codes.ok = 200
         request_mock.post.return_value = post_json_mock
 
         post_data = {
@@ -189,6 +193,8 @@ class OIDCAuthenticationBackendTestCase(TestCase):
             'id_token': 'id_token',
             'access_token': 'access_granted'
         }
+        post_json_mock.status_code = 200
+        request_mock.codes.ok = 200
         request_mock.post.return_value = post_json_mock
         post_data = {
             'client_id': 'example_id',
@@ -245,6 +251,8 @@ class OIDCAuthenticationBackendTestCase(TestCase):
             'id_token': 'token',
             'access_token': 'access_token'
         }
+        post_json_mock.status_code = 200
+        request_mock.codes.ok = 200
         request_mock.post.return_value = post_json_mock
         self.backend.authenticate(request=auth_request)
         calls = [
@@ -276,6 +284,8 @@ class OIDCAuthenticationBackendTestCase(TestCase):
             'id_token': 'token',
             'access_token': 'access_token'
         }
+        post_json_mock.status_code = 200
+        request_mock.codes.ok = 200
         request_mock.post.return_value = post_json_mock
         calls = [
             call('token', 'client_secret', algorithms=['HS256'])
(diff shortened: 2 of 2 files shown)


mozilla_django_oidc/auth.py
```diff
@@ -126,7 +126,9 @@ def authenticate(self, **kwargs):
         response = requests.post(self.OIDC_OP_TOKEN_ENDPOINT,
                                  data=token_payload,
                                  verify=import_from_settings('OIDC_VERIFY_SSL', True))
-        response.raise_for_status()
+
+        if not response.status_code == requests.codes.ok:
+            return None
```
willkg: I'm on the fence about this. What are the situations in which we get a non-ok status code?

* user is denied -- failing the login is ok here
* service is down -- failing the login is ok here, but as a user, I'd really like some indication that the problem wasn't my fault and what I should do about it (probably wait 10 minutes and try again)
* configuration is wrong? -- failing the login is ok here, too, but as a user, I'd really like some indication that the problem wasn't my fault and what I should do about it (probably write up a bug)

Are there other scenarios where this could return a non-ok code?

Are there ways we can get a message to the user so they know what to do next?

willkg: Also, just so I don't forget, any kind of message back to the user requires localization and might be something the site wants to personalize itself.

willkg:
This PR seems to have stalled. I had some thoughts in the comments that I think should get addressed, but otherwise I'm not sure what the next steps are.

mozilla_django_oidc/auth.py
peterbe: Returning `None` is not useful. I agree with @willkg there are many reasons why this might happen. 

Any response error code `>=500` definitely needs to bubble up. If *your* server can't talk to auth0's server it's definitely not the users' fault. Arguably it's not your fault either, but auth0's. 

A huge improvement would be something like:
```python
if response.status_code >= 500:
    raise RemoteTokenEndpointError(response.status_code)
```

peterbe: @willkg [mentioned](https://github.com/mozilla/mozilla-django-oidc/issues/154) that it's happened at least 9 times that he gets a 403 from the OIDC_OP_TOKEN_ENDPOINT. What's that about? 

It would be a small improvement just to raise a specific error just for that but I'd like to know what it means to get a 403 there there. Ideally, it should redirect to a page if possible. E.g.
```python
if response.status_code == 403:
    # That usually means ...
    if settings.import_from_settings('OIDC_403_PAGE_URL', None):
       return redirect(settings.OIDC_403_PAGE_URL)  # perhaps some use of reverse()
    else:
        raise ForbiddenOPError(response)
```

peterbe:
@johngian Why close this? I think it's still a valid concern. 

There are at least two things we can do.

1. Be more patient if you get 5xx errors. E.g. [backoff](https://pypi.org/project/backoff/)
2. Raise an exception if the OIDC provider returns a 403 since that's an urgent indicator that the relationship with the OIDC provider and the RP is sour and needs to be fixed.

johngian:
I think i accidentally deleted the branch and it automatically closed the PR.

peterbe:
Well, I quite like it the way it works now with the `response.raise_for_status()`. It's better than trying to hide failure in the HTTP talk between my server and auth0's server. 

However, perhaps we should split up the work and make a PR just about retries on 5xx errors. And another PR about raising very explicit exceptions if the HTTP communication between the RP and the OIDC provider fails.

akatsoulas:
Closing this due to inactivity. I will also go through the linked issue to see if it's still relevant. @johngian feel free to open a new one if you have some time.

PR #157, opened 2017-06-30
Add note about platform support

willkg:
This adds the trove classifiers for OSX and Linux, but not Windows.

This adds a note to the installation chapter about supported Python versions and
platforms.

Does this look ok? I wondered whether to be explicit and say "We don't support Windows." or not. Would that be a better thing to have in the installation chapter?


akatsoulas:
I believe it's OK to list only the supported platforms and not explicitly saying that we don't support Windows.

willkg:
Thank you!

PR #161, opened 2017-07-14
Implement RS256 verification

### LINKED ISSUE #149: Feature request: support RS256 as signing algorithm
mozilla-django-oidc cannot be used with the provider [Dataporten](https://auth.dataporten.no/.well-known/openid-configuration) as it does not support RS256.

Currently you wind up with a 500 error due to the following exception being raised:

```
JWSError at /oidc/callback/

The specified alg value is not allowed
```

puiterwijk:
Fixes #149

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -6,7 +6,7 @@ import requests
 
 from django.utils.encoding import smart_bytes, smart_text
 from django.contrib.auth import get_user_model
-from django.core.exceptions import SuspiciousOperation
+from django.core.exceptions import SuspiciousOperation, ImproperlyConfigured
 from django.core.urlresolvers import reverse
 
 from jose import jws
@@ -45,6 +45,11 @@ class OIDCAuthenticationBackend(object):
         self.OIDC_OP_USER_ENDPOINT = import_from_settings('OIDC_OP_USER_ENDPOINT')
         self.OIDC_RP_CLIENT_ID = import_from_settings('OIDC_RP_CLIENT_ID')
         self.OIDC_RP_CLIENT_SECRET = import_from_settings('OIDC_RP_CLIENT_SECRET')
+        self.OIDC_RP_SIGN_ALGO = import_from_settings('OIDC_RP_SIGN_ALGO', 'HS256')
+        self.OIDC_RP_IDP_SIGN_KEY = import_from_settings('OIDC_RP_IDP_SIGN_KEY', None)
+
+        if self.OIDC_RP_SIGN_ALGO.startswith('RS') and self.OIDC_RP_IDP_SIGN_KEY is None:
+            raise ImproperlyConfigured('IDP Signing key not provided with RS signing algorithm')
 
         self.UserModel = get_user_model()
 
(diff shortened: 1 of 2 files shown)


puiterwijk:
This will actually implement RS$keysize for any $keysize that python-jose supports, but it will at least support RS256.

johngian:
Thanks for the contribution @puiterwijk

mozilla_django_oidc/auth.py
```diff
@@ -45,6 +45,11 @@ def __init__(self, *args, **kwargs):
         self.OIDC_OP_USER_ENDPOINT = import_from_settings('OIDC_OP_USER_ENDPOINT')
         self.OIDC_RP_CLIENT_ID = import_from_settings('OIDC_RP_CLIENT_ID')
         self.OIDC_RP_CLIENT_SECRET = import_from_settings('OIDC_RP_CLIENT_SECRET')
+        self.OIDC_RP_SIGN_ALGO = import_from_settings('OIDC_RP_SIGN_ALGO', 'HS256')
+        self.OIDC_RP_IDP_SIGN_KEY = import_from_settings('OIDC_RP_IDP_SIGN_KEY', None)
```
johngian: Why do you prefer to have a different config entry for the key? Can't we reuse ``OIDC_RP_CLIENT_SECRET`` and have a string for HS-based algorithms and a jwk dict for RS-based?

johngian:
It's not clear to me which is the preferred format for handling credentials in various OIDC providers. I've only seen PEM. From what I understand you can convert an RSA key to JWK. Can you please update the docs with how to do this or even add some helper code to handle PEM?

mozilla_django_oidc/auth.py
puiterwijk: No, because you still need a CLIENT_SECRET that is different from the signing key.
The client secret that is provided is used for exchanging an authorization code to an access token in the code flow, but the id_token is signed with a different (provider) key.

puiterwijk:
On the preferred key type: the OpenID Connect specs only list JWKs, and all providers that implement the Discovery extension will publish a JWK set (http://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata, e.g. for Fedora at https://id.fedoraproject.org/openidc/Jwks).
The jwks_uri is also mentioned in the core OpenID Connect specifications: http://openid.net/specs/openid-connect-core-1_0.html#RotateSigKeys

It is technically possible indeed to transform keys from PEM to JWK, but I would personally suggest that we just have people insert JWKs since that's what the specs say.

mozilla_django_oidc/auth.py
johngian: Sounds good, thanks for the clarification.

willkg:
Where's this PR at? What's the next step for moving it forward?

puiterwijk:
@willkg I'm basically waiting for @johngian or someone else to look at my code again and my responses to the questions asked.
If you guys, contrary to specification, want PEM keys instead of JWKSets, I can make that change, but in my opinion that would be making things harder for users.

willkg:
@puiterwijk I re-read the thread here. Last comment from @johngian is this one:

https://github.com/mozilla/mozilla-django-oidc/pull/161#issuecomment-316436568

where he asks this:

> Can you please update the docs with how to do this or even add some helper code to handle PEM?

Is that done, yet? If not, then I think that's where things are at.

puiterwijk:
@willkg Right. I was wondering whether the PEM -> JWK doc is still useful if JWK(Set) is the official way to do things, but if you still want that I'll add instructions.

johngian:
Closing this one after #182

PR #162, opened 2017-07-14
Request the correct OpenIDC scopes

puiterwijk:
This patch makes the requested set of OpenID Connect scopes configurable, but
defaults to "openid email".

The "email" scope was never requested, which means that any privacy-sensitive
IdP would not send the users' email address.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -97,6 +97,12 @@ of ``mozilla-django-oidc``.
       When using a custom callback view, it is generally a good idea to subclass the
       default ``OIDCAuthenticationCallbackView`` and override the methods you want to change.
 
+.. py:attribute:: OIDC_RP_SCOPES
+
+   :default: ``openid email``
+
+   The OpenID Connect scopes to request during login.
+
 .. py:attribute:: LOGIN_REDIRECT_URL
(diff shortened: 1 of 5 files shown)


johngian:
Hi @puiterwijk 
Thanks for the contribution!

mozilla_django_oidc/utils.py
```diff
@@ -39,3 +39,19 @@ def is_authenticated(user):
     if _less_than_django_1_10:
         return user.is_authenticated()
     return user.is_authenticated
+
+
+def get_oidc_scopes():
```
johngian: Instead of increasing the complexity by adding an extra function that handles the scopes why don't we just have a config entry with the scopes predefined as a string?

puiterwijk: The main reason for that was to allow people to define the scopes as a list or a string. But if you prefer to just have users configure a string, that works for me.

johngian: Well i don't see the value of having a list of strings and I prefer the less complicated approach.

johngian:
Thanks @puiterwijk 🚀

PR #164, opened 2017-07-26
Remove references to runtests.py.

akatsoulas:


willkg:
@peterbe ^^^ You did the tox overhaul. What do you think the Makefile rules should look like?

PR #165, opened 2017-08-21
Update auth.py

GermanoGuerrini:
VERIFY_SSL option use extended to user endpoint request.


peterbe:
I don't think this warrants a release but it would be right to make a mention of it in the HISTORY. I can do that.

PR #166, opened 2017-08-31
OIDC_STORE_ACCESS_TOKEN doesn't (and isn't documented)

djmitche:
https://github.com/mozilla/mozilla-django-oidc/blob/02502ca99926493649beeab7a82cb8c8fc5144c2/mozilla_django_oidc/auth.py#L137-L138

This setting is undocumented, and seems to store the id_token, rather than the access token.  Is there a reason for that?  If not, do you mind a PR to fix it?

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -103,6 +103,18 @@ of ``mozilla-django-oidc``.
 
    The OpenID Connect scopes to request during login.
 
+.. py:attribute:: OIDC_STORE_ACCESS_TOKEN
+
+   :default: ``False``
+
+   If true, store the access_token in ``session['oidc_access_token']``.
+
+.. py:attribute:: OIDC_STORE_ID_TOKEN
+
+   :default: ``False``
+
+   If true, store the id_token in ``session['oidc_id_token']``.
+
 .. py:attribute:: LOGIN_REDIRECT_URL
 
    :default: ``/accounts/profile``
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -135,6 +135,9 @@ class OIDCAuthenticationBackend(object):
             access_token = token_response.get('access_token')
 
             if import_from_settings('OIDC_STORE_ACCESS_TOKEN', False):
+                session['oidc_access_token'] = access_token
+
+            if import_from_settings('OIDC_STORE_ID_TOKEN', False):
                 session['oidc_id_token'] = id_token
 
             user_response = requests.get(self.OIDC_OP_USER_ENDPOINT,
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -126,6 +126,8 @@ class OIDCAuthenticationBackendTestCase(TestCase):
             headers={'Authorization': 'Bearer access_granted'}
         )
 
+    @override_settings(OIDC_STORE_ACCESS_TOKEN=True)
+    @override_settings(OIDC_STORE_ID_TOKEN=True)
     @patch('mozilla_django_oidc.auth.requests')
     @patch('mozilla_django_oidc.auth.OIDCAuthenticationBackend.verify_token')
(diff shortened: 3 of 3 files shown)


djmitche:
Well, there, I made a PR out of it!  I added `OIDC_STORE_ID_TOKEN` too because why not.

djmitche:
..now with tests :)

docs/settings.rst
```diff
@@ -103,6 +103,18 @@ of ``mozilla-django-oidc``.
 
    The OpenID Connect scopes to request during login.
 
+.. py:attribute:: OIDC_STORE_ACCESS_TOKEN
+
+   :default: ``False``
+
+   If true, store the access_token in ``session['oidc_access_token']``.
```
peterbe: I think this needs work. This sentence only says what it does which might be a good start but it really ought to flesh out a bit more about *why* and *why not* this is something you'd want to set.

docs/settings.rst
```diff
@@ -103,6 +103,18 @@ of ``mozilla-django-oidc``.
...
+
+.. py:attribute:: OIDC_STORE_ID_TOKEN
+
+   :default: ``False``
+
+   If true, store the id_token in ``session['oidc_id_token']``.
```
peterbe: Ditto for the same point above.

mozilla_django_oidc/auth.py
```diff
@@ -135,6 +135,9 @@ def authenticate(self, **kwargs):
             access_token = token_response.get('access_token')
 
             if import_from_settings('OIDC_STORE_ACCESS_TOKEN', False):
+                session['oidc_access_token'] = access_token
+
+            if import_from_settings('OIDC_STORE_ID_TOKEN', False):
```
peterbe: I don't have deep enough insight (partially blame memory) into the core of this. But this seems like a huge difference.

peterbe:
Before this lands we need to review the git commit message to be more descriptive. 
Also, when this lands it looks like it needs a major version bump since the logic about what to store totally changes. 

Awaiting action from @akatsoulas @johngian

djmitche:
I don't think this changes what to store.  It documents an existing feature, and adds a new parallel feature that is not enabled by default.

The reason to want either of these tokens is to use them for other purposes beyond identifying the user to Django.

peterbe:
@djmitche Prior to your change [the code](https://github.com/mozilla/mozilla-django-oidc/blob/7982bd544d68816fde9f9331ae5a251ee5b2f477/mozilla_django_oidc/auth.py#L137-L138) would use `OIDC_STORE_ACCESS_TOKEN` to decide whether to store the `id_token` in the session. Seems like a real oversight. Either it should have been `OIDC_STORE_ID_TOKEN` or `access_token` there to match.

johngian:
* Regarding the config entry name. Indeed it is an oversight that we never really realized it was wrong.
* Regarding storing the tokens
  * We need to allow users to store the ``id_token`` in the session by changing this config entry. That means the config entry name change is valid. 
  * Regarding access tokens, I don't see the point why we need to store them. After sending the request to the OP to get the user info, the access token is invalid and useless for the rest of the AUTHN/AUTHZ procedure.

djmitche:
The auth token isn't invalid until it expires.  What makes you think otherwise?

djmitche:
I think what you're saying is that the only specific use for an access token in the OIDC standard is to call `/userinfo`.  That doesn't mean it's the only *possible* use for that token, and this option makes it available to users for those other purposes.  For the use-case I have in mind, that will mean calling another API with that access token (where that other API is listed among the `aud`iences for the token).

johngian:
@djmitche you are right. I thought that access tokens are revoked after using them. I tested it locally and I could use the same access token multiple times.

johngian:
I updated the docs and squashed the changes to the initial commit here:
https://github.com/mozilla/mozilla-django-oidc/pull/174

johngian:
Closing this one after #174

PR #167, opened 2017-08-31
Allow configuration of additional auth parameters

djmitche:
The audience parameter is included in the `aud` claim in the resulting
tokens, and can allow other RPs to authenticate the tokens as well.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -124,6 +124,12 @@ of ``mozilla-django-oidc``.
    Controls whether the OpenID Connect client stores the OIDC ``id_token`` in the user session.
    The session key used to store the data is ``oidc_id_token``.
 
+.. py:attribute:: OIDC_AUTH_REQUEST_EXTRA_PARAMS
+
+   :default: `{}`
+
+   Additional parameters to include in the initial authorization request.
+
 .. py:attribute:: LOGIN_REDIRECT_URL
 
    :default: ``/accounts/profile``
--- a/mozilla_django_oidc/views.py
+++ b/mozilla_django_oidc/views.py
@@ -133,6 +133,9 @@ class OIDCAuthenticationRequestView(View):
             'state': state,
         }
 
+        extra = import_from_settings('OIDC_AUTH_REQUEST_EXTRA_PARAMS', {})
+        params.update(extra)
+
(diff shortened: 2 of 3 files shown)


peterbe:
@akatsoulas @johngian Can you please review and when you merge, I think this warrents a minor version bump since it's actually a new feature.

johngian:
Hey @djmitche 

I took a look at the PR and code looks OK. That said i cannot find this parameter in the OIDC spec:
http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

Can you point me to some documentation about that parameter?

djmitche:
Hm, maybe it's an Auth0-specific thing?
https://auth0.com/docs/api-auth/tutorials/verify-access-token

Dustin

johngian:
I would prefer to avoid adding custom auth0 functionality since we are aiming to have an OIDC implementation.

That said, we are trying to facilitate our (Mozilla) needs first. Do you think that's reasonable to make it a bit more generic and add a dictionary under settings that users can override with all kinds of values the OIDC authentication init params?

eg:
```
OIDC_RP_AUTH_CUSTOM_PARAMS = {
    'foo': 'bar'
}
```

and do something like ``params.update(OIDC_RP_AUTH_CUSTOM_PARAMS)`` here ?
https://github.com/mozilla/mozilla-django-oidc/pull/167/files#diff-5eb980a040a9c8e409d0e67b94023074R138

djmitche:
Ah, it *is* in the OIDC spec -
http://openid.net/specs/openid-connect-core-1_0.html#IDToken

So I'd argue we should keep the feature as-is.

johngian:
@djmitche This list describes the information that the token returned from the OP after successfully authenticating contains. The parameters to initialize the authentication process are only those listed here:

http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

djmitche:
Ah, good point.  Perhaps we should call this something like "AUTHENTICATION_REQUEST_EXTRA_PARAMS"?

johngian:
For code sanity reasons we are prefixing ``mozilla-django-oidc`` settings with ``OIDC`` so something like ``OIDC_AUTH_REQUEST_EXTRA_PARAMS`` should be OK.

djmitche:
Updated and rebased.

PR #169, opened 2017-09-05
Add test that specs out multi-user return behavior

claudijd:
This is a unit-test addition to verify the behavior of what would happen if we had two users with the same name.

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -80,6 +80,14 @@ class OIDCAuthenticationBackendTestCase(TestCase):
         user = User.objects.create_user('example_username')
         self.assertEqual(self.backend.get_user(user.pk), user)
 
+    def test_get_multiple_user(self):
+        """Test get_user method with multiple valid users."""
+
+        user = User.objects.create_user('example_username')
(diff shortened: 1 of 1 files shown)


willkg:
This adds a test for a situation that can't happen due to db constraints--you can't have two users with the same username.

claudijd:
@willkg yes, I was just realizing that, sort of trial by fire on this one.  The failure behavior is expected, so I'm closing this PR.

willkg:
No worries! I apologize for being overly curt.

I haven't checked, but it'd be good if we had a test for what happens when two users have the same email address--that's the thing we're using to connect an OIDC identity with a Django identity. That's similar to this and maybe what you were trying to get at. Does that sound right?

claudijd:
@willkg yes, that is what I was trying to test.  I will probably send a PR for that soon.  It seems to be a decent testing strategy, assuming I'm not spamming you guys too much.

claudijd:
FWIW, I checked and a test for this does exist...

https://github.com/mozilla/mozilla-django-oidc/blob/master/tests/test_auth.py#L388-L411

PR #170, opened 2017-09-05
Add test for case-mismatch duplicate emails

claudijd:
This just added a case insensitive match test case to provide coverage for case insensitive matching found in filter_users_by_claims method.


tests/test_auth.py
```diff
@@ -409,3 +409,31 @@ def test_duplicate_emails(self, request_mock, jws_mock):
...
+    @patch('mozilla_django_oidc.auth.jws.verify')
+    @patch('mozilla_django_oidc.auth.requests')
+    def test_duplicate_emails_case_mismatch(self, request_mock, jws_mock):
+        """Test auth with two users having the same email, with different case."""
+        auth_request = RequestFactory().get('/foo', {'code': 'foo',
+                                                     'state': 'bar'})
+        auth_request.session = {}
+
+        User.objects.create(username='user1', email='email@example.com')
+        User.objects.create(username='user2', email='eMaIl@ExAmPlE.cOm')
```
claudijd: This is the only delta besides the test case name.

peterbe:
I second that.

peterbe:
Note; this does not require a version bump or release.

PR #171, opened 2017-09-05
Add path host injection unit-test

claudijd:
Adds a check to ensure absolutify cannot be tricked into going to an alternate address.


peterbe:
I second that.

peterbe:
Note; this does not require a version bump or release.

PR #174, opened 2017-09-08
Revisit OIDC_STORE_{ACCESS,ID}_TOKEN config entries

johngian:
* Fix name issue with ``OIDC_STORE_ID_TOKEN`` (was ``OIDC_STORE_ACCESS_TOKEN``)
* Implement functionality to store ``access_token`` in user session and add
  config entry to control this feature

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -103,6 +103,27 @@ of ``mozilla-django-oidc``.
 
    The OpenID Connect scopes to request during login.
 
+.. py:attribute:: OIDC_STORE_ACCESS_TOKEN
+
+   :default: ``False``
+
+   Controls whether the OpenID Connect client stores the OIDC ``access_token`` in the user session.
+   The session key used to store the data is ``oidc_access_token``.
+
+   By default we want to store as few credentials as possible so this feature defaults to ``False``
+   and it's use is discouraged.
+
+   .. warning::
+      This feature stores authentication information in the session. If used in combination with Django's
+      cookie-based session backend, those tokens will be visible in the browser's cookie store.
+
+.. py:attribute:: OIDC_STORE_ID_TOKEN
+
+   :default: ``False``
+
+   Controls whether the OpenID Connect client stores the OIDC ``id_token`` in the user session.
+   The session key used to store the data is ``oidc_id_token``.
+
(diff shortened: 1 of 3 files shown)


johngian:
@djmitche can you double check the changes that i squashed to your commit? If you agree i can merge it.

johngian:
FYI I added a disclaimer about the security issues that might appear when storing access tokens in the session.

docs/settings.rst
```diff
@@ -103,6 +103,29 @@ of ``mozilla-django-oidc``.
...
    The OpenID Connect scopes to request during login.
 
+.. py:attribute:: OIDC_STORE_ACCESS_TOKEN
+
+   :default: ``False``
+
+   Controls whether the OpenID Connect client stores the OIDC ``access_token`` in the user session.
+   The session key used to store the data is ``oidc_access_token``.
+
+   By default we want to store as less credentials as possible so this feature defaults to ``False``
```
djmitche: ..as few..

docs/settings.rst
```diff
@@ -103,6 +103,29 @@ of ``mozilla-django-oidc``.
...
+   and is discouraged to use.
```
djmitche: ..and its use is discouraged.

docs/settings.rst
```diff
@@ -103,6 +103,29 @@ of ``mozilla-django-oidc``.
...
+
+   .. danger::
+      Use at your own risk. By enabling this feature you are storing OIDC credentials to the user session.
+      Don't combine it with Django's cookie-based session backend because this will have as a result leaking
+      OIDC credentials to the client side.
```
djmitche: I think "Use at your own risk" is a little strong here.  The client side is better known as the "user agent" so it's not the worst place to have "user credentials"!  In fact, there are OIDC flows where these tokens are *only* stored on the user agent.  The rest of the note is accurate, so perhaps reword to:

> .. warning::
>     This feature stores authentication information in the session.  If used in combination with Django's [cookie-based session backend](https://docs.djangoproject.com/en/1.11/topics/http/sessions/#using-cookie-based-sessions), those tokens will be visible in the browser's cookie store.

johngian:
Done. @djmitche r?

PR #182, opened 2017-10-02
Implement RS256 verification

### LINKED ISSUE #149: Feature request: support RS256 as signing algorithm

johngian:
Fixes #149

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -130,6 +130,19 @@ of ``mozilla-django-oidc``.
 
    Additional parameters to include in the initial authorization request.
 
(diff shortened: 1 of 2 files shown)


johngian:
Tested this with an auth0 account using both:

* RS256
* HS256

It worked fine for both algorithms

Issue #177, opened 2017-09-19
[docs] Do not suggest using OIDC Auth backend along with ModelBackend


glogiotatidis:
The installation example in the documentation lists both ModelBackend and OIDC backend

```
# Add 'mozilla_django_oidc' authentication backend
AUTHENTICATION_BACKENDS = (
    # ...
    'django.contrib.auth.backends.ModelBackend',
    'mozilla_django_oidc.auth.OIDCAuthenticationBackend',
    # ...
)
```

While this works, I believe that's a wrong security paradigm. If you have enabled OIDC you probably don't want ModelBackend enabled as well, as this can leave an open door for attacks outside the org's SSO system.

I suggest that docs urge users to only use the OIDC backend when enabled.

To allow users to only use OIDC please fix  #176 too

PR #184, opened 2017-10-06
Do not suggest using ModelBackend in installation docs.

johngian:


johngian:
@glogiotatidis r?

johngian:
Fixes #177

docs/installation.rst
```diff
@@ -66,7 +66,6 @@ Start by making the following changes to your ``settings.py`` file.
    # Add 'mozilla_django_oidc' authentication backend
    AUTHENTICATION_BACKENDS = (
        # ...
```
glogiotatidis: I'd remove those `...` at the start and bottom and I'd add some text explaining that it's suggested that you only keep `ModelBackend`

PR #208, opened 2018-01-23
get_extra_params and OIDC_AUTHENTICATE_CLASS

johnpaulett:
I am using Auth0 and need a way to pass the `?connection=` parameter to hint to Auth0 which Identity Provider to use (of which there will be multiple).  

`OIDC_AUTH_REQUEST_EXTRA_PARAMS` would work if I only had one connection option, but I need to dynamically decide which connection param I want to use in the extra params.

Therefore, I have implemented two hooks to allow this behavior:
* `OIDCAuthenticationRequestView.get_extra_params` provides the `OIDC_AUTH_REQUEST_EXTRA_PARAMS`, but allows a subclass to override and dynamically change based upon the `request`
* `OIDC_AUTHENTICATE_CLASS` (modeled after `OIDC_CALLBACK_CLASS`) to allow registering a custom subclass as the view to use for the url.  I debated the naming on this, but picked it to match the `/authenticate/` URL, so as to avoid confusion as being the Authentication backend class.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -97,6 +97,18 @@ of ``mozilla-django-oidc``.
       When using a custom callback view, it is generally a good idea to subclass the
       default ``OIDCAuthenticationCallbackView`` and override the methods you want to change.
 
+.. py:attribute:: OIDC_AUTHENTICATE_CLASS
+
+   :default: ``mozilla_django_oidc.views.OIDCAuthenticationRequestView``
+
+   Allows you to substitute a custom class-based view to be used as OpenID Connect
+   authenticate URL.
+
+   .. note::
+
+      When using a custom authenticate view, it is generally a good idea to subclass the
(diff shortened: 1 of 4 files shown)


tpazderka:
Hi, is this something that has a chance to get into master?

We need this functionality for our deployment.

If the original author cannot update this PR, I am willing to do that.

willkg:
The code changes here look fine to me. This should have a test to make sure it works as advertised and also to help us know if we break it in the future through other changes.

@johnpaulett Can you add a test to this PR? If not, let me know and I can write one up.

johnpaulett:
@willkg just saw you're request about tests -- will do.

johnpaulett:
Test added to cover the custom get_extra_params. Let me know if you see anything else that needs to be covered.

akatsoulas:
Since this is already approved I am merging it. Thanks @johnpaulett and @willkg!

PR #209, opened 2018-02-09
Update user claims

johngian:


mozilla_django_oidc/auth.py
```diff
@@ -203,7 +207,7 @@ def authenticate(self, **kwargs):
             users = self.filter_users_by_claims(user_info)
 
             if len(users) == 1:
-                return users[0]
+                return self.update_user(users[0], user_info)
```
peterbe: How about:
```python
if len(users) == 1:
    self.update_user(users[0], user_info)
    return users[0]
```
...since it's mutable.

akatsoulas:
@peterbe is this ready to be merged?

johnpaulett:
I struggled with a better name as well.  Tried to mimic the create_user method since it returns the user object. 

Let me know if I need to do anything to get it merged in.

peterbe:
@akatsoulas Perhaps it got lost in the communication, but I think this is ready to be merged.

PR #211, opened 2018-03-16
support ajax-y sites, fixes #133

### LINKED ISSUE #133: support AJAX-y sites
The current middleware and the new one don't support AJAX-y sites.

This issue covers figuring out how web-sites that are predominantly doing AJAX requests to move data around will work with this library.

peterbe:
I'm not super proud of this code but I think it's a lot better than what we had before. Before, if you have a Django authentication, which then starts a single page app thing that does a bunch of XHR requests only, eventually you'll be kicked out. This patch makes it possible to do something about that. 

One thing I think is a bit weird is that the `JsonResponse` comes with a header AND a JSON blob with a key called `refresh_url`. Only one of them gets used really. Headers are usually better but also, it's a bit unusual with a header key with a `_` in it.


peterbe:
I don't think we need to make a release just for this. We'll have to remember to mention it in the CHANGELOG on the next release. If you prefer, perhaps make a release anyway so we don't forget :)

PR #212, opened 2018-03-16
used cached_property on exempt_urls

peterbe:
One thing that hurt my eye when I looked through the code is that we compute a list of URLs on every single request. Even when the list is never even used. 
It's pretty much CPU-bound so it wasn't horribly slow but it hurt to see it go to waste.

This PR makes the computation of all `exempt_urls` a once-per-processor thing. 
It also changes to a set which is faster to look for an item in. 
Lastly it rearranges the order so the comparison of `request.path` with the exempt URLs happens last.

### DIFF
--- a/mozilla_django_oidc/middleware.py
+++ b/mozilla_django_oidc/middleware.py
@@ -14,6 +14,7 @@ except ImportError:
     from django.core.urlresolvers import reverse
 from django.http import HttpResponseRedirect
 from django.utils.crypto import get_random_string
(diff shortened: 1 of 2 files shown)


mozilla_django_oidc/middleware.py
```diff
@@ -59,10 +62,10 @@ def get_exempt_urls(self):
             'oidc_logout',
         ])
 
-        return [
+        return set([
```
willkg: Good idea! +1

Issue #153, opened 2017-06-19
/oidc/callback/ infinite redirect loop


willkg:
I switched standups from our home-grown auth0 client implementation to mozilla-django-oidc 0.3.1 on Friday. At least one user went to use the site and immediately entered an infinite redirect on this url:

```
https://www.standu.ps/oidc/callback/?error=login_required&error_description=Login%20required&state=EYy2Qz8peNMCImxor3RKHGAGf8ggKLvh
```

Clearing the browser's standup cookies fixed it.

It's possible that it's the `/oidc/callback/` view does its thing and there's one branch that redirects to itself or something. Maybe this is a problem in conjunction with the middleware?

willkg:
I have a few users who keep hitting the infinite redirect, so I don't think the theory that it was a transition issue holds water.

Maybe what's going on is that the browser goes to `/oidc/callback` with an error. The callback view has no explicit handling for this. The callback view redirects the user to the failure url which by default is `/`. Standup is using the middleware, so then the middleware processes that request and sends the browser back to Mozilla SSO which sends the browser back to `/oidc/callback` with the same error. Repeat.

That sure seems like what's going on.

Assuming that's true, I think to fix this we should:

1. add handling in `/oidc/callback` for when the querystring params have `error` and `error_description` and redirect to the failure url carrying that error code along
2. the failure url should not default to `/`--it should be an explicit view for "Yo--you fail at login." messages
3. the failure url should be exempted by the middleware

@johngian @akatsoulas @peterbe Does that sound right/likely?

pmac:
Seeing this on basket as well.

peterbe:
`error` and `error_description` aren't particular oidc specific so you couldn't do something like `if request.GET.get('error') and request.GET.get('error_description')` in the middleware. 

I too think the right thing to do is to have a another URL to redirect to that is not `/` and then the RP is responsible for making such a page. And the middleware should then do something like `if settings.RP_ERROR_URL == request.path: return`.

peterbe:
This is likely to be very related: https://github.com/mozilla/airmozilla/issues/827#issuecomment-374274903
What happened there was that I was logged in (thank you django session cookies) but since my `oidc_id_token_expiration` has expired so mozilla_django_oidc redirects me to `/authorize?...&prompt=none` which then errors out and redirects back to `/oidc/callback` which redirects to `/login-failure/` (which is a public django page) and there the middleware kicks in and redirects back again. 

[...]

PR #213, opened 2018-03-19
Log out when session refresh fails

peterbe:
I admit that I did this quite hastily. But the principle is sound. If the callback is unhappy with the client, the client shouldn't be allowed to stay logged in. Not just in principle but because if you're allowed to stay logged in, when you reach the login_failure URL (whatever that is) the middleware might think it's time to refresh your session and you get swept up in a loop.

### DIFF
--- a/mozilla_django_oidc/views.py
+++ b/mozilla_django_oidc/views.py
@@ -63,7 +63,16 @@ class OIDCAuthenticationCallbackView(View):
             # Make sure that nonce is not used twice
             del request.session['oidc_nonce']
 
-        if 'code' in request.GET and 'state' in request.GET:
+        if request.GET.get('error'):
+            # Ouch! Something important failed.
+            # Make sure the user doesn't get to continue to be logged in
+            # otherwise the refresh middleware will force the user to
+            # redirect to authorize again if the session refresh has
+            # expired.
+            if is_authenticated(request.user):
+                auth.logout(request)
+            assert not is_authenticated(request.user)
+        elif 'code' in request.GET and 'state' in request.GET:
             kwargs = {
                 'request': request,
                 'nonce': nonce,
--- a/tests/test_middleware.py
+++ b/tests/test_middleware.py
@@ -308,7 +308,7 @@ class MiddlewareTestCase(TestCase):
             'error_description': 'Multifactor authentication required',
         })
         self.assertEqual(resp.status_code, 302)
-        # Note, in other versions of Django this 'resp.url' will be
+        # Note, in versions of Django <=1.8, this 'resp.url' will be
         # an absolute URL, so we need to make this split to make sure the
         # test suite works in old and new versions of Django.
         if 'http://testserver' in resp.url:
--- a/tests/test_middleware.py
+++ b/tests/test_middleware.py
@@ -308,7 +308,13 @@ class MiddlewareTestCase(TestCase):
             'error_description': 'Multifactor authentication required',
         })
         self.assertEqual(resp.status_code, 302)
-        self.assertEquals(resp.url, '/')
+        # Note, in other versions of Django this 'resp.url' will be
+        # an absolute URL, so we need to make this split to make sure the
+        # test suite works in old and new versions of Django.
+        if 'http://testserver' in resp.url:
+            self.assertEquals(resp.url, 'http://testserver/')
+        else:
+            self.assertEquals(resp.url, '/')
 
         # Since the user in 'client' doesn't change, we have to use other
         # queues to assert that the user got logged out properly.
--- a/tests/test_middleware.py
+++ b/tests/test_middleware.py
@@ -11,8 +11,10 @@ from mock import patch
 import django
 from django.conf.urls import url
 from django.contrib.auth import get_user_model
+from django.contrib.auth.signals import user_logged_out
 from django.contrib.auth.models import AnonymousUser
 from django.core.cache import cache
+from django.dispatch import receiver
 from django.http import HttpResponse
 from django.test import Client, RequestFactory, TestCase, override_settings
 from django.test.client import ClientHandler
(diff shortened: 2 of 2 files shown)


tests/test_middleware.py
```diff
@@ -258,3 +260,69 @@ def test_expired_token_redirects_to_sso(self, mock_random_string):
...
+        )
+        # Now suppose the user goes there and something goes wrong.
+        # For example, the user might have become "blocked" or the 2FA
+        # verficiation has expired and needs to be done again.
+        resp = client.get('/callback/', {
+            'error': 'login_required',
+            'error_description': 'Multifactor authentication required',
+        })
+        self.assertEqual(resp.status_code, 302)
+        # Note, in other versions of Django this 'resp.url' will be
```
willkg: When you say "other versions of Django", what're you talking about? Do you mean pre-2.0 versions?

tests/test_middleware.py
```diff
@@ -258,3 +260,69 @@ def test_expired_token_redirects_to_sso(self, mock_random_string):
...
+        # an absolute URL, so we need to make this split to make sure the
+        # test suite works in old and new versions of Django.
+        if 'http://testserver' in resp.url:
+            self.assertEquals(resp.url, 'http://testserver/')
+        else:
+            self.assertEquals(resp.url, '/')
```
willkg: I'd probably do:

```
self.assertIn(resp.url, ('/', 'http://testserver/'))
```

Either is fine, but it's shorter.

tests/test_middleware.py
```diff
@@ -258,3 +260,69 @@ def test_expired_token_redirects_to_sso(self, mock_random_string):
...
+
+        # Since the user in 'client' doesn't change, we have to use other
+        # queues to assert that the user got logged out properly.
+
+        # The session gets flushed when you get signed out.
+        # This is the only decent way to know the user lost all
+        # request.session and
+        self.assertTrue(not client.session.items())
```
willkg: You can do `assertFalse` here.

tests/test_middleware.py
peterbe: `<=1.8` I believe.

mozilla_django_oidc/views.py
```diff
@@ -63,7 +63,16 @@ def get(self, request):
...
-        if 'code' in request.GET and 'state' in request.GET:
+        if request.GET.get('error'):
+            # Ouch! Something important failed.
+            # Make sure the user doesn't get to continue to be logged in
+            # otherwise the refresh middleware will force the user to
+            # redirect to authorize again if the session refresh has
+            # expired.
+            if is_authenticated(request.user):
+                auth.logout(request)
+            assert not is_authenticated(request.user)
```
willkg: I think this sounds ok. In https://github.com/mozilla/mozilla-django-oidc/issues/153#issuecomment-317597945 I suggested we add a failure url that's exempt from the middleware and redirect to that.

I don't know offhand if the error message that comes from the SSO is always user-friendly or not. I wonder if it's always the case that an error from the SSO indicates that the user should get logged out of the Django site? Are there any other possible situations?

tests/test_middleware.py
peterbe: Neat!

tests/test_middleware.py
peterbe: Perhaps enough pytest has entered my blood veins that I'm thinking of `assert not something_true`. Also, there is no other `self.assertFalse` usage.

willkg: I'm totally with you. I only knew about it because I looked up `assertEquals`. I thought there was only `assertEqual`. Fun fun fun!

mozilla_django_oidc/views.py
peterbe: Yeah, this is tricky. First of all, I can't talk about if the SSO ever redirects to `yoursite/oidc/callback/?error=something` where the error isn't "major". 

An alternative solution is to do something like this in the RefreshIDToken middleware:

```python
if request.path == self.failure_url() and request.GET.get('error'):
    # Don't redirect to refresh the ID token! Something's up!
else:
   # the existing regular stuff
```

The counter-argument is that if Auth0 does redirect to the callback with an error it sounds pretty severe and correctly the user should perhaps be kicked out. In my solution, what would happen from a user perspective is "What? I was logged in last Friday and now I no longer am. Oh well, I'll just log in again" and when you do that if Auth0 has an issues they'll be dealt with there. In the case of the Airmozilla failure it seemed the error was that [I hadn't 2FA'ed in a while](https://github.com/mozilla/airmozilla/issues/827#issuecomment-374274903).

Currently, if you try go to `https://www.standu.ps/oidc/callback/?error=something&error_message=Blablabla` you're just redirected back to `https://www.standu.ps/` and that `error` ad `error_message` strings are lost. [Any interesting query string is discarded.](https://github.com/mozilla/mozilla-django-oidc/blob/da3d49a275c1f7bfba49a1f6b9d017090a566325/mozilla_django_oidc/views.py#L83)

peterbe: Also, suppose we had a way of making a dedicated "OIDC callback error page" where RPs can localize the content and potentially display (escaped!) the `error` and `error_message` strings then what will the user do with it? Read it? 
The user just wants to log in and if he was already logged in, he'd have to click the sign in button again. Or sign out. And if the sign in page (e.g. `https://air.mozilla.org/authentication/signin/`) is outside that exempt scheme mentioned above the redirect would start again because of the RefreshIDToken middleware.

peterbe: Would really appreciate some feedback here @johngian @akatsoulas

akatsoulas: I think the simplest solution would be to just logout the user. Displaying the error message probably won't be that useful to the user. Moreover, for me it makes sense to logout the user if there is a failure in the refresh process

peterbe: @akatsoulas Is that an approval? :)

akatsoulas: I didn't do a review because @willkg did all the hard work here. Logging the user out sounds good to me

tests/test_middleware.py
willkg: Let's fix the comment here to say `<=1.8` if that's what it's about.

peterbe:
@willkg Comment fixed. 

Remind me, what's the policy for this project in terms of who merges and stuff?

willkg:
I don't know what the policy is, but I don't merge things in this project.

akatsoulas:
I believe the author merges the PR when it's approved.

peterbe:
@akatsoulas Kind sir, can you please make a release now? I think this feature is big enough to be worthy of a new release.

akatsoulas:
@johngian just released this.

Issue #204, opened 2017-12-14
Exception when calling /oidc/callback


nfvs:
Using Microsoft's servers triggers this exception:
```
Traceback (most recent call last):
  File "/Users/nsantos/.virtualenvs/ngn-console/lib/python3.6/site-packages/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/Users/nsantos/.virtualenvs/ngn-console/lib/python3.6/site-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/Users/nsantos/.virtualenvs/ngn-console/lib/python3.6/site-packages/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/nsantos/.virtualenvs/ngn-console/lib/python3.6/site-packages/django/views/generic/base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "/Users/nsantos/.virtualenvs/ngn-console/lib/python3.6/site-packages/django/views/generic/base.py", line 88, in dispatch
    return handler(request, *args, **kwargs)
  File "/Users/nsantos/.virtualenvs/ngn-console/lib/python3.6/site-packages/mozilla_django_oidc/views.py", line 75, in get
    self.user = auth.authenticate(**kwargs)
  File "/Users/nsantos/.virtualenvs/ngn-console/lib/python3.6/site-packages/django/contrib/auth/__init__.py", line 70, in authenticate
    user = _authenticate_with_backend(backend, backend_path, request, credentials)
  File "/Users/nsantos/.virtualenvs/ngn-console/lib/python3.6/site-packages/django/contrib/auth/__init__.py", line 115, in _authenticate_with_backend
    return backend.authenticate(*args, **credentials)
  File "/Users/nsantos/.virtualenvs/ngn-console/lib/python3.6/site-packages/mozilla_django_oidc/auth.py", line 174, in authenticate
    if self.verify_token(id_token, nonce=nonce):
  File "/Users/nsantos/.virtualenvs/ngn-console/lib/python3.6/site-packages/mozilla_django_oidc/auth.py", line 123, in verify_token
    smart_bytes(key),
  File "/Users/nsantos/.virtualenvs/ngn-console/lib/python3.6/site-packages/mozilla_django_oidc/auth.py", line 95, in _verify_jws
    if not jws.verify(jwk):
[...]

PR #223, opened 2018-04-23
fail earlier when algorithm does not match

anlutro:
Currently you get a pretty obsucre error if the algorithm set by the provider doesn't match what the client thinks should be used. https://github.com/mozilla/mozilla-django-oidc/issues/204

By checking the algorithm before trying to verify the JWS token, hopefully the error message will be a little more clear.

I was considering adding a test for this but it seems like all of the existing tests mock `_verify_jws` so I wasn't sure.

Let me know if the new error message is correct.

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -99,18 +99,21 @@ class OIDCAuthenticationBackend(ModelBackend):
         """Verify the given JWS payload with the given key and return the payload"""
 
         jws = JWS.from_compact(payload)
-        jwk = JWK.load(key)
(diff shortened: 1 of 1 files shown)


johngian:
I verified it locally using a client configured to use ``HS256`` in the provider level and ``OIDC_RP_SIGN_ALGO`` set to ``RS256`` and code works as expected.

johngian:
Code looks OK. Thanks @anlutro for the contribution. Closing this one after #228 
:rocket:

anlutro:
Thanks @johngian !

PR #225, opened 2018-04-24
Change RefreshIDToken to SessionRefresh, fixes #200

### LINKED ISSUE #200: Change variable name `mozilla_django_oidc.middleware.RefreshIDToken` to something else
Kind of a nitpick, however, `RefreshIDToken` is a confusing name because refresh tokens are a very specific thing: non-expiring access tokens.
In this case, this is the opposite, its a token that expires and needs periodic refresh.

willkg:
The middleware has nothing to do with refresh tokens, so switching the
name makes this less confusing.


willkg:
This is a backwards incompatible API change, so I upped the version to 1.0.0.

HISTORY.rst
```diff
@@ -3,14 +3,15 @@
 History
 -------
 
-0.7.0 (unreleased)
+1.0.0 (unreleased)
```
johngian: ``0.7.0`` is not released so we don't need to bump another version since its already backwards incompatible.

johngian:
Code looks OK. I am not sure about the version bump though, why do we need to change it to `1.0.0` ?

HISTORY.rst
willkg: If we're doing semver, then this is a backwards incompatible API change which requires a major version bump.

Issue #222, opened 2018-04-20
OIDCAuthenticationBackend captures authentication which doesn't belong to it


ziima:
Calling `authenticate(request=request, my_kwarg=my_value)` with `OIDCAuthenticationBackend` in `AUTHENTICATION_BACKENDS` results in `SuspiciousOperation`. `OIDCAuthenticationBackend` shouldn't care about authentication, if the arguments doesn't match it's signature.

ziima:
Fixed by #226. Thanks

PR #226, opened 2018-04-24
Return `None` when authenticate() request has no `state` or `code`

johngian:
In order for the OIDCAuthenticationBackend to be combined with other auth
backends it needs to return `None` when the call arguments are not targeting
this specific backend. This way we allow Django to try the next available
backend defined in settings.


johngian:
Fixes #222

johngian:
@willkg r?

Issue #227, opened 2018-04-27
Is it possible to use mozilla-django-oidc together with Django REST framework?


NorthCat:
Is it possible to use mozilla-django-oidc together with Django REST framework?
If so, please give detailed instructions.

anlutro:
I'm pretty sure this is not currently possible, but it would be a great addition.

Here's the DRF docs on how to implement an authentication backend: http://www.django-rest-framework.org/api-guide/authentication/#custom-authentication

Basically you just need an `authenticate(self, request)` method which returns the user model + arbitrary(?) authentication information based on the request (by looking at tokens in headers or whatever).

I think this just means extracting [this block of code](https://github.com/mozilla/mozilla-django-oidc/blob/c6d9906aa24a549ee9313023e638b4e1e6bda64a/mozilla_django_oidc/auth.py#L197-L228) into a standalone method/function, which can then be re-used in the DRF authentication backend.

johngian:
Fixed after #231

PR #179, opened 2017-09-23
Refactor auth backend

woutor:
Fixes #149, #178

Just introduced some methods to easily extend the `mozilla_django_oidc.auth.OIDCAuthenticationBackend`

PR #231, opened 2018-05-08
implement authentication class for DRF

anlutro:
Related issue: https://github.com/mozilla/mozilla-django-oidc/issues/227

Mostly this involves refactoring the part that transforms a request (specifically its `Authorization` header) into a user object, so that there's a method you can use for *only* retrieving the user, which can then be re-used in the stateless/session-less DRF authentication class.

While working on this I found some minor issues which I extraced into smaller individual PRs. This PR depends on these, and I will rebase this PR when/if those are merged:

- [x] ~#233~ #241 
- [x] #234

### DIFF
--- a/mozilla_django_oidc/contrib/drf.py
+++ b/mozilla_django_oidc/contrib/drf.py
@@ -26,24 +26,22 @@ def get_oidc_backend():
     # allow the user to force which back backend to use. this is mostly
     # convenient if you want to use OIDC with DRF but don't want to configure
     # OIDC for the "normal" Django auth.
-    backend_setting = import_from_settings('OIDC_AUTH_BACKEND', None)
+    backend_setting = import_from_settings('OIDC_DRF_AUTH_BACKEND', None)
     if backend_setting:
         backend = import_string(backend_setting)()
         if not isinstance(backend, OIDCAuthenticationBackend):
-            msg = 'Class configured in OIDC_AUTH_BACKEND ' \
+            msg = 'Class configured in OIDC_DRF_AUTH_BACKEND ' \
                   'does not extend OIDCAuthenticationBackend!'
             raise ImproperlyConfigured(msg)
         return backend
 
     # if the backend setting is not set, look through the list of configured
     # backends for one that is an OIDCAuthenticationBackend.
-    backends = [
-        backend for backend in get_backends()
-        if isinstance(backend, OIDCAuthenticationBackend)
-    ]
+    backends = [b for b in get_backends() if isinstance(b, OIDCAuthenticationBackend)]
+
     if not backends:
         msg = 'No backends extending OIDCAuthenticationBackend found - ' \
-              'add one to AUTHENTICATION_BACKENDS or set OIDC_AUTH_BACKEND!'
+              'add one to AUTHENTICATION_BACKENDS or set OIDC_DRF_AUTH_BACKEND!'
         raise ImproperlyConfigured(msg)
     if len(backends) > 1:
         raise ImproperlyConfigured('More than one OIDCAuthenticationBackend found!')
--- a/mozilla_django_oidc/contrib/drf.py
+++ b/mozilla_django_oidc/contrib/drf.py
@@ -1,0 +1,132 @@
+"""
+Classes/functions for integrating with Django REST Framework.
+
+http://www.django-rest-framework.org/api-guide/authentication/#custom-authentication
+"""
+
+import logging
+
+from django.contrib.auth import get_backends
+from django.core.exceptions import ImproperlyConfigured, SuspiciousOperation
+from django.utils.module_loading import import_string
+from rest_framework import authentication, exceptions
+from requests.exceptions import HTTPError
+
+from mozilla_django_oidc.auth import OIDCAuthenticationBackend
+from mozilla_django_oidc.utils import import_from_settings, parse_www_authenticate_header
+
+LOGGER = logging.getLogger(__name__)
+
+
+def get_oidc_backend():
+    """
+    Get the Django auth backend that uses OIDC.
+    """
+
+    # allow the user to force which back backend to use. this is mostly
+    # convenient if you want to use OIDC with DRF but don't want to configure
+    # OIDC for the "normal" Django auth.
+    backend_setting = import_from_settings('OIDC_AUTH_BACKEND', None)
+    if backend_setting:
+        backend = import_string(backend_setting)()
+        if not isinstance(backend, OIDCAuthenticationBackend):
+            msg = 'Class configured in OIDC_AUTH_BACKEND ' \
+                  'does not extend OIDCAuthenticationBackend!'
+            raise ImproperlyConfigured(msg)
+        return backend
+
+    # if the backend setting is not set, look through the list of configured
+    # backends for one that is an OIDCAuthenticationBackend.
+    backends = [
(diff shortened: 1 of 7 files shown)


johngian:
Hey @anlutro ! Thanks for the contribution.
This PR needs rebase after #179 got merged.

anlutro:
@johngian done. I'm curious if my approach looks good. I needed to change `get_or_create_user` so it didn't try to mess with sessions, and also to throw `SuspiciousOperation` (maybe it should throw something else) so that the DRF code can differentiate between authentication failure and an error.

kerrermanisNL:
Nice, would be great to have this! :taco:

anlutro:
Hmm, I just realized that by initializing `OIDCAuthenticationBackend` we prevent people from overriding stuff like `update_user`.

johngian:
Hey @anlutro, thanks for your contribution!

Some high level comments:

* Can you please squash the PR to one or more atomic commits
* I would prefer to maintain the drf support under contrib since it's not something related to the core functionality

mozilla_django_oidc/drf.py
```diff
@@ -0,0 +1,104 @@
...
+                raise exceptions.AuthenticationFailed(data['error_description'])
+
+            # for all other http errors, just re-raise the exception.
+            raise
+        except SuspiciousOperation as exc:
+            # trust that the message in the exception is presentable to the user
+            raise exceptions.AuthenticationFailed('Login failed: %s' % exc)
+
+        if not user:
+            raise exceptions.AuthenticationFailed('Login failed: No user found '
```
johngian: Let's avoid multiline strings. I would prefer:
```
msg = 'Login failed: No user found for given access token'
raise exceptions.AuthenticationFailed('msg')

**Comment by anlutro** [comment]
Rebased. I left off one commit which was a somewhat unrelated refactoring, but can squash that as well if you prefer.

**Comment by johngian** [comment]
It looks like i submitted the change requests for the version before you rebase. Can you take a look at the outdated comments that target ``mozilla_django_oidc/drf.py``?

If that's not convenient i can always copy/paste the old comments to the current codebase after your last changes.

**Comment by anlutro** [comment]
No need, I'll reply or fix them as is :)

File: mozilla_django_oidc/drf.py [hunk 2]
```diff
@@ -0,0 +1,104 @@
...
+        raise ImproperlyConfigured('more than one OIDCAuthenticationBackend found!')
+    return backends[0]
+
+
+class OIDCAuthentication(authentication.BaseAuthentication):
+    """
+    Provide OpenID authentication for DRF.
+    """
+
+    www_authenticate_realm = 'api'
```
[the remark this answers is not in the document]
anlutro: I added some comments to clarify this. Mostly I just copied it from https://github.com/encode/django-rest-framework/blob/26b0f650d622dbbc2cf8d8b38a9c52b5b0d89860/rest_framework/authentication.py#L61 - if DRF doesn't find it useful to define as a setting, I guess we shouldn't either.

mozilla_django_oidc/drf.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def __init__(self, backend=None):
```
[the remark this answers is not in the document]
anlutro: I don't think there is a way to configure DRF to pass in the backend manually. The primary purpose of it here is honestly just to make it easier to set it up for automated testing. I don't think it'll cause any harm.

anlutro:
Fixed some things, replied to some others

mozilla_django_oidc/contrib/drf.py
```diff
@@ -0,0 +1,110 @@
...
+class OIDCAuthentication(authentication.BaseAuthentication):
+    """
+    Provide OpenID authentication for DRF.
+    """
+
+    # used by the authenticate_header method.
+    www_authenticate_realm = 'api'
+
+    def __init__(self, backend=None):
+        self.backend = backend or guess_oidc_backend()
```
johngian: Is there any case that we don't have OIDC enabled for the django app itself but only for the drf component? That way we imply that we use the authentication backend. Let's add a configuration entry to manually configure backend but also try to autoconfigure.

mozilla_django_oidc/contrib/drf.py
```diff
@@ -0,0 +1,110 @@
...
+            # we can get from the www-authentication header) in the response.
+            if resp.status_code == 401 and 'www-authenticate' in resp.headers:
+                data = parse_www_authenticate_header(resp.headers['www-authenticate'])
+                raise exceptions.AuthenticationFailed(data['error_description'])
+
+            # for all other http errors, just re-raise the exception.
+            raise
+        except SuspiciousOperation as exc:
+            # trust that the message in the exception is presentable to the user
+            raise exceptions.AuthenticationFailed('Login failed: %s' % exc)
```
johngian: I would prefer to avoid sending the actual exception to the end user response. Let's add a generic message there or just leave only the error code.

mozilla_django_oidc/contrib/drf.py
```diff
@@ -0,0 +1,110 @@
...
+        return user, access_token
+
+    def get_access_token(self, request):
+        """
+        Get the access token based on a request.
+
+        Returns None if no authentication details were provided. Raises
+        AuthenticationFailed if the token is incorrect.
+        """
+        header = authentication.get_authorization_header(request).decode('ascii')
```
johngian: It looks like ``get_authorization_header`` returns a string encoded in ``iso-8859-1``. Why are you decoding to ``ascii``?

https://github.com/encode/django-rest-framework/blob/master/rest_framework/authentication.py#L27

mozilla_django_oidc/contrib/drf.py
```diff
@@ -0,0 +1,110 @@
...
+        if not header:
+            return None
+
+        auth = header.split()
+
+        if auth[0].lower() != 'bearer':
+            return None
```
johngian: What is the reasoning behind return ``None`` if not authentication details are provided but raise exception in the rest of the failures?

johngian:
@anlutro i added a couple of more comments but after that it looks like PR is ready to be merged.

mozilla_django_oidc/contrib/drf.py
anlutro: Good point, I've added a specific setting for this.

mozilla_django_oidc/contrib/drf.py
anlutro: There is no error code, but I'll change it to just a generic "Login failed" and log the actual exception (without a stacktrace) instead.

mozilla_django_oidc/contrib/drf.py
anlutro: I guess I assumed it would always be ascii. Changed to use the constant which DRF uses.

mozilla_django_oidc/contrib/drf.py
anlutro: If `None` is returned here, then `authenticate()` also returns `None`, which means DRF won't return a 401 (which would imply that credentials were provided but were incorrect).

anlutro:
Amended with changes. By the way, if you want to do any final touchups yourself, I've checked "Allow edits from maintainers", so you should be able to push directly to my PR branch without having to open a completely new PR.

johngian:
:rocket:

PR #232, opened 2018-05-15
Renew `oidc_login_next` after user login

### LINKED ISSUE #224: OIDCAuthenticationCallbackView.success_url ignores session value
Because `OIDCAuthenticationCallbackView.success_url` is called only after `auth.login`, the session is already cycled and user is redirected according to settings instead of URL stored session.

tpazderka:
Close #224

### DIFF
--- a/mozilla_django_oidc/views.py
+++ b/mozilla_django_oidc/views.py
@@ -46,7 +46,9 @@ class OIDCAuthenticationCallbackView(View):
         return HttpResponseRedirect(self.failure_url)
 
     def login_success(self):
+        next_url = self.request.session.get('oidc_login_next', None)
         auth.login(self.request, self.user)
+        self.request.session['oidc_login_next'] = next_url
 
         # Figure out when this id_token will expire. This is ignored unless you're
         # using the RenewIDToken middleware.
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -8,8 +8,9 @@ from mock import patch
 
 import django
 from django.core.exceptions import SuspiciousOperation
-from django.contrib.auth import get_user_model
+from django.contrib.auth import get_user_model, login
 from django.contrib.auth.models import AnonymousUser
+from django.contrib.sessions.backends.cache import SessionStore
 try:
     from django.urls import reverse
 except ImportError:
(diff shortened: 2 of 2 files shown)


dmptrluke:
Is there any way this can be merged? I'm having to keep a fork for this bug fix.

TCFox:
Still eagerly awaiting this fix, as it's causing a number of headaches. There's a fork being maintained for this fix at https://github.com/dmptrluke/mozilla-django-oidc ([diff](https://github.com/mozilla/mozilla-django-oidc/compare/master...dmptrluke:master)) but I would love to see this resolved here as it would make a lot of things easier.

Is there anything I can offer to assist with merging this PR?

akatsoulas:
I am not sure I completely understand this patch. My understanding is that the goal of this PR is to allow some custom logic to populate the `oidc_login_next` variable. If this is the case, wouldn't overriding the `success_url` suffice in this case?

This is a long standing PR and the logic has changed a bit since this was opened. I am closing this, mainly because of the reason I gave upon. Please do reopen if this is not case.

akatsoulas:
Following up on this, the only case that the data in the `oidc_login_next` can be overridden is in the `get_next_url`. Is this the actual problem? Having a hook for a custom `get_next_url` method could solve for this.

mariadb-BryanAlsdorf:
Not the author, but this is a bug fix to handle users not being redirect to their original destination under certain circumstances.

When calling `auth.login(self.request, self.user)`, in some circumstances [django resets the session for security purposes](https://github.com/django/django/blob/main/django/contrib/auth/__init__.py#L106).

With the session reset, `oidc_login_next` (set it https://github.com/mozilla/mozilla-django-oidc/blob/master/mozilla_django_oidc/middleware.py#L161) is no longer available so the user is not redirected back to the original destination.

It has been a while since I had to work around this, but the main circumstance that trigger this was when the end user logs out and then logs back in with an active session. I also believe it can be triggered when the OP and RP have different session lengths, but I don't recall all the specifics off hand.

akatsoulas:
Thank you both @mariadb-BryanAlsdorf @tpazderka . I was mainly confused by the issue that was focusing on the login process and I completely missed the middleware.

PR #233, opened 2018-05-17
move session management from get_or_create_user to authenticate

anlutro:
This is something which was introduced in https://github.com/mozilla/mozilla-django-oidc/pull/179.

I think it's wrong to do anything with the session in a `get_or_create_user` method.

Hopefully this shouldn't affect anyone overriding the method(s).

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -239,14 +239,13 @@ class OIDCAuthenticationBackend(ModelBackend):
         verified_id = self.verify_token(id_token, nonce=nonce)
 
         if verified_id:
+            self._write_tokens_to_session(access_token, id_token, verified_id)
             return self.get_or_create_user(access_token, id_token, verified_id)
 
         return None
 
-    def get_or_create_user(self, access_token, id_token, verified_id):
-        """Returns a User instance if 1 user is found. Creates a user if not found
-        and configured to do so. Returns nothing if multiple users are matched."""
-
+    def _write_tokens_to_session(self, access_token, id_token, verified_id):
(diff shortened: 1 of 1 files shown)


anlutro:
Tests seem to fail on Travis because `wheel` stopped supporting Python 3.3?

johngian:
@anlutro Tests should be fixed after #239 can you rebase ?

anlutro:
@johngian done!

johngian:
Please change the commit name to something more specific since we are not actually handling the session management in general but specifically storing the OIDC tokens.

johngian:
r+wc

mozilla_django_oidc/auth.py
```diff
@@ -247,15 +255,6 @@ def get_or_create_user(self, access_token, id_token, verified_id):
         """Returns a User instance if 1 user is found. Creates a user if not found
         and configured to do so. Returns nothing if multiple users are matched."""
 
-        session = self.request.session
-
-        if import_from_settings('OIDC_STORE_ACCESS_TOKEN', False):
```
johngian: Ideally i would prefer we move storing tokens to another method since it's quite self contained.

anlutro:
Amended to "write oidc tokens to session in authenticate(), not get_or_create_user()"

mozilla_django_oidc/auth.py
anlutro: Sounds reasonable, will do that.

anlutro:
Extracted a `_write_tokens_to_session` method.

johngian:
Closing this one after #241.

Thanks @anlutro for the contribution :rocket:

PR #234, opened 2018-05-17
extract username generation to an overridable method

anlutro:
Currently you can override `username_algo` but that assumes that you want to generate the username based on the email. This is not always the case.

Adding a method `get_username` lets us generate usernames based on `sub`, for example.

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -77,22 +77,26 @@ class OIDCAuthenticationBackend(ModelBackend):
 
     def create_user(self, claims):
         """Return object for a newly created user account."""
-        # bluntly stolen from django-browserid
-        # https://github.com/mozilla/django-browserid/blob/master/django_browserid/auth.py
 
-        username_algo = import_from_settings('OIDC_USERNAME_ALGO', None)
         email = claims.get('email')
         if not email:
             return None
 
+        username = self.get_username(claims)
+
+        return self.UserModel.objects.create_user(username, email)
+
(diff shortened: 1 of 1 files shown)


anlutro:
Tests seem to fail on Travis because `wheel` stopped supporting Python 3.3?

mozilla_django_oidc/auth.py
```diff
@@ -77,22 +77,25 @@ def verify_claims(self, claims):
 
     def create_user(self, claims):
         """Return object for a newly created user account."""
-        # bluntly stolen from django-browserid
```
johngian: Let's also keep the reference in ``get_username``.

mozilla_django_oidc/auth.py
```diff
@@ -77,22 +77,25 @@ def verify_claims(self, claims):
...
         email = claims.get('email')
         if not email:
             return None
 
+        username = self.get_username(claims)
+
+        return self.UserModel.objects.create_user(username, email)
+
+    def get_username(self, claims):
+        """Get the local username based on claims."""
```
johngian: I am not sure if ``local username`` makes a lot of sense. I would prefer something like:

```Generate username based on claims.```

johngian:
Thanks @anlutro for the PR. Code looks really straightforward and it doesn't cause any incompatibilities. I added a couple of comments regarding comments/docstrings.

anlutro:
Feedback incorporated

johngian:
Closing this one after #242

Issue #249, opened 2018-06-11
Unsecured JWT check is failing


ollamh:
# Problem
We have an identity provider which doesn't sign tokens (and to my surprise it's legal `https://tools.ietf.org/html/rfc7519#section-6`), so, the token looks like
```
<header>.<payload>.
```
and this causing the error:
```
  File "/var/proj/virtual/local/lib/python2.7/site-packages/mozilla_django_oidc/auth.py", line 239, in authenticate
    verified_id = self.verify_token(id_token, nonce=nonce)
  File "/var/proj/virtual/local/lib/python2.7/site-packages/mozilla_django_oidc/auth.py", line 165, in verify_token
    verified_token = self._verify_jws(token, key)
  File "/var/proj/virtual/local/lib/python2.7/site-packages/mozilla_django_oidc/auth.py", line 103, in _verify_jws
    jws = JWS.from_compact(payload)
  File "/var/proj/virtual/local/lib/python2.7/site-packages/josepy/jws.py", line 313, in from_compact
    signature=b64.b64decode(signature))
  File "/var/proj/virtual/local/lib/python2.7/site-packages/josepy/jws.py", line 166, in __init__
    kwargs = self._with_combined(kwargs)
  File "/var/proj/virtual/local/lib/python2.7/site-packages/josepy/jws.py", line 177, in _with_combined
    combined = header + cls.header_cls.json_loads(protected)
  File "/var/proj/virtual/local/lib/python2.7/site-packages/josepy/interfaces.py", line 179, in json_loads
    return cls.from_json(loads)
  File "/var/proj/virtual/local/lib/python2.7/site-packages/josepy/json_util.py", line 289, in from_json
    return cls(**cls.fields_from_json(jobj))
  File "/var/proj/virtual/local/lib/python2.7/site-packages/josepy/json_util.py", line 280, in fields_from_json
    fields[slot] = field.decode(value)
  File "/var/proj/virtual/local/lib/python2.7/site-packages/josepy/json_util.py", line 88, in decode
    return self.fdec(value)
  File "/var/proj/virtual/local/lib/python2.7/site-packages/josepy/jwa.py", line 57, in from_json
    return cls.SIGNATURES[jobj]
KeyError: u'none'
```
# Expected behavior
IMHO, the token should pass the check (since no signing information has been given) and maybe log a warning about the token not being signed. Or maybe it would be even better to have a flag which can allow/forbid usage of unsigned tokens.

[...]

PR #251, opened 2018-06-13
Allow using unsecured JWT token

johngian:
* Add config entry OIDC_VERIFY_JWS that defaults to True
* Change phrasing to allow the notion of unsecured tokens


johngian:
Fixes #249

johngian:
@akatsoulas r?

What happens here is that the spec allows JWT that have no signature if the header has ``{"alg": "none"}``. I added some code and config to handle this case and a lot of tests around verifying JWTs. can you take a look ?

tests/test_auth.py
```diff
@@ -75,6 +79,134 @@ def test_invalid_token(self, request_mock, token_mock):
...
+            smart_text(b64encode(header)),
+            smart_text(b64encode(payload)),
+            signature
+        ))
+
+        with self.assertRaises(KeyError):
+            self.backend.get_payload_data(token, None)
+
+    @override_settings(OIDC_ALLOW_UNSECURED_JWT=True)
+    def test_allowed_unsecured_secure_token_secure_verified(self):
```
akatsoulas: this name is really confusing.

tests/test_auth.py
```diff
@@ -75,6 +79,134 @@ def test_invalid_token(self, request_mock, token_mock):
...
+            smart_text(b64encode(payload)),
+            smart_text(signature)
+        )
+        token_bytes = force_bytes(token)
+        key_text = smart_text(key)
+        output = self.backend.get_payload_data(token_bytes, key_text)
+        self.assertEqual(output, payload)
+
+    @override_settings(OIDC_ALLOW_UNSECURED_JWT=False)
+    def test_disallowed_unsecured_secure_token_secure_verified(self):
```
akatsoulas: same here

akatsoulas: the same goes for the other tests too. I would suggest to tweak it a bit so someone can understand easily what the test is doing.

mozilla_django_oidc/auth.py
```diff
@@ -153,6 +154,15 @@ def retrieve_matching_jwk(self, token):
             raise SuspiciousOperation('Could not find a valid JWKS.')
         return key
 
+    def get_payload_data(self, token, key):
+        """Helper method to get the payload of the JWT token."""
+        if import_from_settings('OIDC_ALLOW_UNSECURED_JWT', False):
```
akatsoulas: Please add a couple of comments in the code to shortly explain the two cases (token verification and unsecured).

johngian: Done a couple of lines later.

PR #256, opened 2018-07-20
Fixes #247 - Makes alg optional in a JWK

### LINKED ISSUE #247: alg should be optional
An error is thrown if the RSA key doesn't have it

https://imgur.com/a/LZiDZ9f

Algogator:


johngian:
Thanks @Algogator for the contribution
Closing this one after introducing a couple of tests in #258 

:rocket:

PR #270, opened 2018-08-09
Fix CI

johngian:


johngian:
@akatsoulas r?

* Adds Django 2.1 to CI
* Removes unsupported python 3.4 add python 3.7

PR #273, opened 2018-08-16
problem: retrieve_matching_jwk breaks when multiple keys are returned

JustinAzoff:
When looping over the keys in retrieve_matching_jwk it needs to first
skip over any key whose kid does not match BEFORE it checks if the alg
matches.  Otherwise at least one of the keys will not match and
SuspiciousOperation will be raised for every response.


JustinAzoff:
Related to https://github.com/mozilla/mozilla-django-oidc/issues/247

johngian:
Closing this one after #274 
Thanks @JustinAzoff for the contribution :rocket:

PR #277, opened 2018-09-11
Adds settings to control redirect behavior

chrisbrantley:
We ran into some issues integrating mozilla-django-oidc because our client application is hosted on a subdomain. We could not redirect back to the previous URL because it was not on the same domain as the django app and therefore failed the `is_safe()` check in `mozilla_django_oidc.views.get_next_url`.

That check requires that the redirect URL be on the same domain (a problem for us) AND that it use HTTPS if the django site uses HTTPS (great for production, not so great for local development when there client app is running on localhost).

So this PR adds 2 new settings: 

OIDC_REDIRECT_REQUIRE_HTTPS=True|False (defaults to the current `request.is_secure()` behavior 
OIDC_REDIRECT_ALLOWED_HOSTS=tuple|list of hostnames

This allows us to whitelist our client app domain and also supports some flexibility during local development. Thank you!

### DIFF
--- a/mozilla_django_oidc/views.py
+++ b/mozilla_django_oidc/views.py
@@ -115,11 +115,13 @@ def get_next_url(request, redirect_field_name):
 
         # NOTE(robhudson): Django 2.1 changes `host` to `allowed_hosts`.
         if django.VERSION >= (2, 1):
-            hosts = tuple(getattr(settings, 'OIDC_REDIRECT_ALLOWED_HOSTS', ())) + (request.get_host(),)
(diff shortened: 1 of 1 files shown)


johngian:
Closing this one after #287. Thank you @chrisbrantley for the contribution. Apparently after #286 changes got more simple. I pushed all the PR fixes and amended your commit here:
https://github.com/mozilla/mozilla-django-oidc/pull/287/commits/5eca37891bbebf376e7669bcb4c62df8071f2c85

:rocket:

PR #278, opened 2018-09-17
Verify email is in claims before creating user

kerrermanisNL:
Previously it could occur that during authentication a user would try to be created without the email claim. However the create_user function would not allow for this, returning `None`. On the surface everything seems to work fine, but no user would actually be created with no debug or warning messages to speak of. Instead of always returning `True` for `verify_claims` it would therefor make more sense if this function actually checked if the email is present before trying to create the user.

If this would break existing functionality perhaps some explicit warning message should be logged when the `create_user` function would not create the user and return `None`. To prevent hours of debugging for people :) 

Haven't updated the `HISTORY.rst` file since I wasn't sure how you guys wanted to have the changes present (and versioning etc). And I saw you guys were the more recent comitters.

Let me know if any changes need to be made to the PR, or if I should go for the logging alternative.

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -74,17 +74,12 @@ class OIDCAuthenticationBackend(ModelBackend):
 
     def verify_claims(self, claims):
         """Verify the provided claims to decide if authentication should be allowed."""
-        return True
+        return 'email' in claims
 
     def create_user(self, claims):
         """Return object for a newly created user account."""
-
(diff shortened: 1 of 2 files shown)


kerrermanisNL:
Hi, anyone willing to review this?

johngian:
Closing this one after #285. Thanks @kerrermanisNL for the contribution.

kerrermanisNL:
Thanks for the merge ;)

PR #300, opened 2019-03-01
Make JWKS KID validation optional with the OIDC_VERIFY_KID setting

fronbasal:
Hey!
I created a patch to make the KID validation optional using a setting.

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -158,7 +158,7 @@ class OIDCAuthenticationBackend(ModelBackend):
 
         key = None
         for jwk in jwks['keys']:
-            if jwk['kid'] != smart_text(header.kid) and import_from_settings("OIDC_VERIFY_KID", True):
+            if import_from_settings("OIDC_VERIFY_KID", True) and jwk['kid'] != smart_text(header.kid):
                 continue
             if 'alg' in jwk and jwk['alg'] != smart_text(header.alg):
                 continue
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -19,7 +19,6 @@ from josepy.jws import JWS, Header
 
 from mozilla_django_oidc.utils import absolutify, import_from_settings
 
-
 LOGGER = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


keenandr:
+1. Not all providers use kid (ex. https://auth.globus.org/jwk.json ) and kid is optional per JWK spec ( https://tools.ietf.org/html/rfc7517#section-4.5 ). If there's a risk with not checking kid, the code could check to make sure kid is not null before using it.

fronbasal:
I think the user should be able to decide whether they want to validate KID or not, even if it not null (if e.g. the OAuth server is misconfigured and delivers an invalid KID).

peterbe:
I haven't worked on this project for a very long time. And I don't know why only Travis checks linting.

fronbasal:
@peterbe All right. Is this project actively maintained? Else I think it would be sad if it were to die in obsolescence because this is good code.

peterbe:
@johngian Do you have time to look at this PR?

wanaryytel:
I'm working with a provider that doesn't provide a `kid` value at all in which case this would result in a `KeyError` (as it does now, too). Maybe this should be considered before merging this in?

fronbasal:
@wanaryytel great case. Unfortunately I don’t think this PR will be merged at all.  Feel free to fork my fork ;)

wanaryytel:
@fronbasal Thanks, I did :) https://github.com/fronbasal/mozilla-django-oidc/pull/1
Not much else I can do but keep your PR up-to-date.

mozilla_django_oidc/auth.py
```diff
@@ -157,7 +157,7 @@ def retrieve_matching_jwk(self, token):
 
         key = None
         for jwk in jwks['keys']:
-            if jwk['kid'] != smart_text(header.kid):
+            if jwk['kid'] != smart_text(header.kid) and import_from_settings("OIDC_VERIFY_KID", True):
```
cfra, line 160: This might be a better fix for #375 instead of #387 if #387 has security implications that are not acceptable.

For this to work, the parts need to be swapped though, because the `KeyError` will still happen otherwise:

```suggestion
            if import_from_settings("OIDC_VERIFY_KID", True) and jwk['kid'] != smart_text(header.kid):
```

cfra (changes requested):

fronbasal:
I'm sorry I haven't looked into this PR for quite a while. I'll rebase it to master and have a look at @cfra's suggestion. Thank you for keeping this alive!

fronbasal:
I think the rebase went well, can someone confirm? 😅

cfra:
I don't think the rebase went as intended.

I am not 100% sure what went wrong there, but looking at "Files changed" for this PR, there seem to be a lot of changes that are not related to your pull request.

To clean this up, I would probably skip the merges and break history using something like:

```console
git reset --hard upstream/master  # Depending how the upstream branch is called in your repository
git cherry-pick <commit that implements the change>
git push --force-with-lease
```

But that's just my 2 cents.

fronbasal:
@cfra Thank you so much for the suggestion, I think it worked out now. Can you confirm?

cfra (approved):
This looks good to me now. :sparkles: 

Now, let's hope the maintainers find the time to take another look at this. :slightly_smiling_face:

fronbasal:
@cfra Thank you so much!

fronbasal:
@johngian Could you take a look at this PR? :)

akatsoulas:
I merged #387 before going through this PR and it caused a conflict. I rebased this PR in #392. Thanks @fronbasal and @cfra for your work on this issue.

PR #301, opened 2019-03-12
Add  RefreshOIDCToken middleware

GermanoGuerrini:
I noticed that the library was missing the ability to use any refresh token issued by the provider.
It can be used in place of a silent user re-authentication (which does not work for the provider I was working with).
Information can be found [here](https://auth0.com/docs/tokens/refresh-token/current).
Now, while the implementation in this pull request works, it misses a proper provider response validation as described in the [OpenID Specs](https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse).
That would require a serious refactoring in order to store extra information from the original response in a  safer place than the user session, like the database, which currently isn't involved at all.
Plus, the validation should be factored out from the OIDCAuthenticationBackend to be re-used wherever needed.
Is there any interest in such contribution?


titilambert:
Any news on that ?

Galtozzy:
Any updates?

PR #310, opened 2019-05-30
Add OIDC_EXEMPT_URL_PREFIXES setting

jwhitlock:
Add a setting ``OIDC_EXEMPT_URL_PREFIXES``, which works like ``OIDC_EXEMPT_URLS``, avoiding session renewal for any request path that starts with an absolute path string.

On Socorro, we suspect that JavaScript requests with an OIDC session renewal causes problems. We've exempted some of these endpoints (see https://github.com/mozilla-services/socorro/pull/4957), but others have a parameter in the URL string, like ``/signature/graph/<field>/``, which takes a field to graph. 

This would allow us to exclude these paths as well:

```
OIDC_EXEMPT_URLS = [
    'signature:signature_summary',
    'signature:signature_reports',
    ...
]
OIDC_EXEMPT_URL_PREFIXES = [
    '/signature/graph/',
    '/signature/aggregation/',
]
```

Others with similar usage may have a consistent path for JavaScript requests, such as ``/api/``, and could use this instead of listing all request paths.

Alternatively, our code could use a setting ``OIDC_EXEMPT_AJAX``, which would avoid session refresh if ``request.is_ajax()``.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,15 @@
 History
 -------
 
+1.2.4 (Unreleased)
+===================
+
+* Add ``OIDC_EXEMPT_URL_PREFIXES``, a list of absolute URL path prefixes which
+  are exempt from session refreshes in ``SessionMiddleware``.
+  Thanks `@jwhitlock`_
+
+.. _`@jwhitlock`: https://github.com/jwhitlock
+
 1.2.3 (2020-01-02)
 ===================
 
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -81,6 +81,14 @@ of ``mozilla-django-oidc``.
    mozilla-django-oidc urls are exempted from the session renewal by the
    ``SessionRefresh`` middleware.
 
+.. py:attribute:: OIDC_EXEMPT_URL_PREFIXES
+
+   :default: ``[]``
+
+   This is a list of absolute url path prefixes, such a ``['/foo/']``.
+   A request that starts with one of these prefixes is exempt from the session
+   renewal by the ``SessionRefresh`` middleware.
+
 .. py:attribute:: OIDC_CREATE_USER
 
    :default: ``True``
--- a/mozilla_django_oidc/middleware.py
+++ b/mozilla_django_oidc/middleware.py
@@ -73,12 +73,14 @@ class SessionRefresh(MiddlewareMixin):
         if backend_session:
             auth_backend = import_string(backend_session)
             is_oidc_enabled = issubclass(auth_backend, OIDCAuthenticationBackend)
+        exempt_prefixes = list(self.get_settings('OIDC_EXEMPT_URL_PREFIXES', []))
 
         return (
             request.method == 'GET' and
             request.user.is_authenticated and
             is_oidc_enabled and
-            request.path not in self.exempt_urls
+            request.path not in self.exempt_urls and
+            not any(request.path.startswith(prefix) for prefix in exempt_prefixes)
         )
 
     def process_request(self, request):
--- a/tests/test_middleware.py
+++ b/tests/test_middleware.py
@@ -239,6 +239,37 @@ class MiddlewareTestCase(TestCase):
             [u'/authenticate/', u'/callback/', u'/foo/', u'/logout/']
         )
 
+    def test_is_refreshable_url(self):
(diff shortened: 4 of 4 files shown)


willkg:
I wonder if it's better to add support for a prefix syntax for the existing setting. Currently, we have a "if it starts with a `/`, it's a url and if not it's a view name". I think we could add a "if it ends with a `*`, then it's a prefix url".

That seems more consistent with the behavior with the existing setting to me and it feels more predictable.

jwhitlock:
I considered that, but I think the natural next request would be for a regex, and then ``/foo/*`` would look like the weird special case. What if we leap ahead to the regex case?

```
OIDC_EXEMPT_URLS = [
    'signature:signature_summary',                 # no leading slash, its a URL name
    '/signature/reports',                          # leading slash, its a URL path
    re.compile(r'/signature/aggregation/.*'),      # regex, use .match
    ...
]

**Comment by willkg** [comment]
What's the usecase for a regex? What would the urls and paths look like such that a regex is necessary?

**Comment by jwhitlock** [comment]
> What's the usecase for a regex? What would the urls and paths look like such that a regex is necessary?

A Django REST Framework style API, where a file extension is used as an alternate to an ``Accept`` header, for example ``/api/v1/posts/1.html`` versus ``/api/v1/posts/1.json``. You'd expect a browser for ``.html``, and require a session refresh check, but not for the ``.json`` or ``.xml``.

```
OIDC_EXEMPT_URLS = [
    re.compile(r'/api/v1/[^/]*/.*\.(json|xml)$')
]
```

A ``request.is_ajax()`` check would be more precise.

**Comment by johngian** [comment]
Codewise looks OK. I don't have a strong opinion between regex or prefix as long as its well documented. I would prefer though the simplest approach as long as it does the job.

**Comment by johngian** [comment]
Regarding the ``is_ajax`` it can be implemented using a different config entry.

**Comment by willkg** [comment]
I think I'd be +1 on a regex match.

If we ever need to get more flexible in the future, we can also support matching functions that take the request in and return a True/False and then we never have to think about this again.

**Comment by jwhitlock** [comment]
@willkg politely told me, that if I just use my Page Down keys, I'd see an ``is_ajax`` handler later in the  middleware code:

https://github.com/mozilla/mozilla-django-oidc/blob/169590640598c8c0df631c25426f332fc67b61eb/mozilla_django_oidc/middleware.py#L135-L146

This seems reasonable for the default middleware that ships with the library.

> I think I'd be +1 on a regex match.

Do you want a new PR that implements prefix matches with a regex?

**Comment by akatsoulas** [comment]
@jwhitlock since this is already approved, do you mind rebasing this so that we can merge it?

**Comment by jwhitlock** [comment]
I'll rebase.

I avoided merging because of the open question - should this be a new setting ``OIDC_EXEMPT_URL_PREFIXES``, or should ``OIDC_EXEMPT_URLS`` accept regexes as well as strings? I think the regex would be more flexible. I may also be the only one in the world with this issue, and should just implement the regex solution 😓.

**Comment by akatsoulas** [comment]
@jwhitlock I don't have strong feelings about this but maybe a slight preference towards the regex. As you mentioned, it feels more flexible and we can avoid adding a new setting (which of course is not a problem). Please let me know if I shall merge this or wait for the regex approach.

**Comment by jwhitlock** [comment]
Thanks, the regular expression pattern version is in PR #360.

**Comment by jwhitlock** [comment]
Replaced by #360. Instead of another setting, ``OIDC_EXEMPT_URLS`` accepts compiled regex patterns, as well as URL names and absolute URLs. Socorro will use it like this:

```
OIDC_EXEMPT_URLS = [
    # Used by supersearch page as an XHR
    "supersearch:search_fields",  # data-fields-url
    "supersearch:search_results",  # data-results-url
    # Used by bugzilla.js
    "/buginfo/bug",
    # Used by signature report as an XHR
    "signature:signature_summary",  # data-urls-summary
    "signature:signature_reports",  # data-urls-reports
    "signature:signature_bugzilla",  # data-urls-bugzilla
    "signature:signature_comments",  # data-urls-comments
    "signature:signature_correlations",  # data-urls-correlations
    re.compile(r"^/signature/graphs/?P<field>\w+)/$"),  # data-urls-graphs
    re.compile(r"^/signature/aggregation/(?P<aggregation>\w+)/$"),  # data-urls-aggregations
]
```

PR #311, opened 2019-05-31
Expand OIDC_EXEMPT_URLS to take prefixes

jwhitlock:
When an absolute URL has a trailing ``*``, such as ``'/api/*'``, treat it as a prefix for URLs, such as ```/api/posts/3```, that should be exempt from session renewal.

This was suggested by @willkg as an alternate approach to PR #310.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,8 +3,16 @@
 History
 -------
 
+1.2.3 (Unreleased)
+++++++++++++++++++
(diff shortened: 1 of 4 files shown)


jwhitlock:
Closing in favor of #310, which has more of the discussion around this setting change.

jwhitlock:
Oops, said I was closing but neglected to hit the "close" button.

PR #318, opened 2019-08-28
Add OIDCClient for django tests

kerrermanisNL:
Add an OIDCClient for django tests. This can be useful when you need to use the `oidc_id_token` in your views. The default Django test client does not provide this. By making use of this new client your requests will contain a `oidc_id_token` key with a random value (which you can specify). Because there is no easy way to do this yourself for the Django test client (you can either fix it by writing a custom client or by using a `RequestFactory`, both are not very ideal) I thought it would be useful to provide a default client that comes with the mozilla library.

### DIFF
--- a/tests/test_oidc_client.py
+++ b/tests/test_oidc_client.py
@@ -1,10 +1,11 @@
+from unittest import TestCase
+
 from django.test.client import Client
-from django.test.testcases import SimpleTestCase
 
 from mozilla_django_oidc.test import OIDCClient
 
 
-class TestOIDCClient(SimpleTestCase):
+class TestOIDCClient(TestCase):
     def test_inherits_from_django_test_client(self):
         self.assertIsInstance(OIDCClient(), Client)
 
--- a/tests/test_oidc_client.py
+++ b/tests/test_oidc_client.py
@@ -1,0 +1,22 @@
+from django.test.client import Client
+from django.test.testcases import SimpleTestCase
+
+from mozilla_django_oidc.test import OIDCClient
+
+
+class TestOIDCClient(SimpleTestCase):
+    def test_inherits_from_django_test_client(self):
+        self.assertIsInstance(OIDCClient(), Client)
+
+    def test_oidc_id_token_is_set_to_some_token_by_default(self):
+        self.assertEqual(OIDCClient.oidc_id_token, 'some_oidc_token')
+
+    def test_sets_oidc_id_token_for_session(self):
+        self.assertEqual(OIDCClient().session['oidc_id_token'], 'some_oidc_token')
+
+    def test_sets_specified_id_token_for_session(self):
+        class StubOIDCClient(OIDCClient):
+            oidc_id_token = 'some_other_oidc_token'
+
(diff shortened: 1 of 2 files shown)


tests/test_oidc_client.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def test_oidc_id_token_is_set_to_some_token_by_default(self):
+        self.assertEqual(OIDCClient.oidc_id_token, 'some_oidc_token')
+
+    def test_sets_oidc_id_token_for_session(self):
+        self.assertEqual(OIDCClient().session['oidc_id_token'], 'some_oidc_token')
+
+    def test_sets_specified_id_token_for_session(self):
+        class StubOIDCClient(OIDCClient):
+            oidc_id_token = 'some_other_oidc_token'
```
kerrermanisNL: Note that I tested it like this because the session property of the django test client is a little weird to wrap your head around. So instead of instantiating and specifying a new oidc token and refreshing the session, I thought a test like this would be better.

kerrermanisNL:
Usage when specifying this class remains the same for the test client, you just have to specify it in the test case:
```
class TestSomething(TestCase):
    client_class = OIDCClient

    def test_calls_something_with_oidc_token(self):
        # Functionality remains exactly the same, except the view does not die when 
        # accessing `request.session['oidc_id_token']`
        self.client.get('/someting/')
```

kerrermanisNL:
Hi there, could someone have a look at this?

kerrermanisNL:
Could someone have a look at this?

kerrermanisNL:
Bump :)

tests/test_oidc_client.py
```diff
@@ -0,0 +1,22 @@
+from django.test.client import Client
+from django.test.testcases import SimpleTestCase
+
+from mozilla_django_oidc.test import OIDCClient
+
+
+class TestOIDCClient(SimpleTestCase):
```
hartwork: I think deriving from `TestCase` may work better since none of the [things added by `SimpleTestCase`](https://docs.djangoproject.com/en/3.0/topics/testing/tools/#django.test.SimpleTestCase) seem needed here — are they?

kerrermanisNL: I've changed it to make use of `unittest.Testcase`. I assume that's what you mean, since the Django `TestCase` class inherits from `SimpleTestCase` :)

hartwork: It's the other way around — see https://github.com/django/django/blob/a6b3938afc0204093b5356ade2be30b461a698c5/django/test/testcases.py#L146 — but thanks for the change!

kerrermanisNL: Yes that's what I meant :) `django.TestCase` inherits from `django.SimpleTestCase` which inherits from `unittest.TestCase` :smile: Anyhoo, no worries, thanks for the feedback :)

gene1wood:
@kerrermanisNL I've removed the set of reviewers you requested review from as that group are the org owners. They don't have context on this `mozilla-django-oidc` repo and don't perform PR reviews in it.

PR #321, opened 2019-10-10
Tweak logout template examples

jaap3:
* Add missing `csrf_token` / `csrf_input`
* Shorten snippets by removing `html`/`body` tags 
* Fix Jinja2 example

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -159,37 +159,29 @@ Django templates example:
 
 .. code-block:: html+django
 
-   <html>
-     <body>
-       {% if user.is_authenticated %}
-         <p>Current user: {{ user.email }}</p>
-         <form action="{% url 'oidc_logout' %}" method="post">
-           <input type="submit" value="logout">
-         </form>
-       {% else %}
-         <a href="{% url 'oidc_authentication_init' %}">Login</a>
-       {% endif %}
-     </body>
-   </html>
-
+   {% if user.is_authenticated %}
+     <p>Current user: {{ user.email }}</p>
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -159,37 +159,29 @@ Django templates example:
...
-         <form action="{% url('oidc_logout') %}" method="post">
-           <input type="submit" value="logout">
-         </form>
-       {% else %}
-         <a href="{{ url('oidc_authentication_init') }}">Login</a>
-       {% endif %}
-     </body>
-   </html>
-
+   {% if request.user.is_authenticated() %}
```
jaap3: `is_authenticated` is a property in recent Django versions (since 1.10) and can not be called since the release of Django 2.0. Should I change this bit of the snippet as well?

akatsoulas: I would say yes. The snippet above has the property version so it will help us be consistent.

docs/installation.rst
```diff
@@ -159,37 +159,29 @@ Django templates example:
...
-       {% else %}
-         <a href="{{ url('oidc_authentication_init') }}">Login</a>
-       {% endif %}
-     </body>
-   </html>
-
+   {% if request.user.is_authenticated %}
+     <p>Current user: {{ request.user.email }}</p>
+     <form action="{{ url('oidc_logout') }}" method="post">
+       {{ csrf_input() }}
```
akatsoulas: Thanks @jaap3 for this PR. I believe the `parentheses` is not needed here. Other than that I believe it's looking good.

PR #327, opened 2019-11-26
docs(DRF): Fix OIDC_DRF_AUTH_BACKEND reference

SpyTec:
There is a typo to an unresolved class on DRF docs, this changes that to point to auth module


johngian:
Hi @SpyTec thanks for the contribution

PR #329, opened 2019-11-28
fix(Middleware): Check settings for callback URL

SpyTec:
There is a `settings.OIDC_AUTHENTICATION_CALLBACK_URL` which is obeyed in `auth.py`. This makes it uniform and redirects to the correct endpoint


akatsoulas:
Thanks @SpyTec :rocket:

PR #331, opened 2019-11-29
Allow get and post logouts

SpyTec:
Right now only POST is allowed when logging out, this also adds GET


willkg:
This is intentional:

https://github.com/mozilla/mozilla-django-oidc/issues/126

I think if you want to support GET requests, you should subclass.

SpyTec:
I see, thanks for linking the issue

PR #322, opened 2019-10-14
Django compat

jaap3:
This removes references to unsupported Django versions and backwards compatible code that is no longer required and adds Django 2.2 to the test matrices.

PR #336, opened 2019-12-23
Django compat

johngian:


johngian:
I tested it locally and lint/unit/integration tests are passing. This PR introduced a couple of fixes based on the initial #322

* Leave the authentication shim because its a breaking change and just remove the Django 1.10 compat code
* Use the new docker images for testing after migrating `mozilla/docker-test-mozilla-django-oidc` to mozilla org
* Fixes circle CI config that was broken in the initial PR and didn't allow builds to run

That also depends on https://github.com/mozilla/docker-test-mozilla-django-oidc/pull/14 because testrp doesn't work with Django 3x.

I think after we merge the testprovider compatibility fixes the checks here are going to be green again.

johngian:
This is blocked by: https://github.com/mozilla/docker-test-mozilla-django-oidc/pull/15

jaap3:
Awesome, thanks to everyone involved 👍

PR #342, opened 2020-01-20
Don't access email field in get_or_create_user

cfra:
Not all id providers allow authentication clients to use the email
scope.

This is especially true under GDPR, when the email is not needed by the
authentication client to provide its services.

The only usage of the email field in `get_or_create_user` is to print a
debug message if an unknown user tries to authenticate and automatic
account creation is disabled.

By abstracting this access by moving it into a new method
`describe_user_by_claims`, authentication clients which identify users
by other claims (e.g. `sub`), can now do so without having to
duplicate large parts of the `get_or_create_user` methods when
subclassing `OIDCAuthenticationBackend`.


mozilla_django_oidc/auth.py
```diff
@@ -321,8 +323,9 @@ def get_or_create_user(self, access_token, id_token, payload):
             user = self.create_user(user_info)
             return user
         else:
-            LOGGER.debug('Login failed: No user with email %s found, and '
-                         'OIDC_CREATE_USER is False', email)
+            LOGGER.debug('Login failed: No user with %s found, and '
+                         'OIDC_CREATE_USER is False',
+                         self.describe_user_by_claims(user_info))
```
akatsoulas, line 328: I agree that having the email in this method is not useful at all. Probably is something that was left over while `get_or_create_user` was evolving. That said, I don't believe we need another method here. As you mentioned, the email field is only used here so this could be just `claims.get('email', '')`

akatsoulas: All valid points. I didn't think this thoroughly. Thanks for taking the time explaining this.

akatsoulas (approved):

akatsoulas:
@cfra could you please rebase this to the latest master?

Issue #303, opened 2019-03-21
'oidc_state' does not match the OIDC callback state


rackymuthu:
I'm run the django project using the Mozilla Django OIDC. While the project running it is shown the **'oidc_state' does not match the OIDC callback state** error rise in console. how to fix

johngian:
Hey @rackymuthu. Can you provide some more information about your dev environment ?

stevejalim:
I'm seeing this intermittently, too, on a Wagtail project using `mozilla-django-oidc`. 

Sometimes, when opening a new page/following a link in the Wagtail CMS UI, it'll 500.

I don't currently have a full traceback, but I screenshotted the top-line details when I managed ti trigger it in local dev. (Fairly usless, sorry.) 

![Screenshot 2019-11-01 at 13 40 53](https://user-images.githubusercontent.com/101457/68115403-3cf7f500-fef0-11e9-8a80-6e99ba6e402f.png)

The project is using the middleware that re-confirms the user's authorisation with the OpenID provider, at the default 15-minute interval. My gut says this _might_ be a factor, but I haven't yet looked at anything proper. 

Will update this ticket if/when I get a full traceback.

Flor1an-dev:
I've created a merge request which should fix this issue: https://github.com/mozilla/mozilla-django-oidc/pull/345

akatsoulas:
We can track this issue through #312 which has more information about this problem.

Issue #312, opened 2019-07-18
JWT Nonce verification failed


cfra:
In my application, I am seeing a `SuspiciousOperation: JWT Nonce verification failed.` in the logs from time to time.

I am not sure what the reason for this is.

The identity provider seems to always return the nonce correctly.

One guess of mine is that these are maybe flaky clients where the first request for the callback page makes it through to the server, but the response which sets up the authenticated session doesn't make it back to the client. When the user then reloads the callback page, the nonce has already been stripped from the session and so verification fails.

Is this a valid and likely assumption? Or is there a more plausible explanation?

Looking at the control flow, it seems like login will fail for users which trigger this path.

There is however also an option `OIDC_USE_NONCE` which would permit to turn off this validation completely.

I couldn't find any documentation about the security implications of that setting. Is there any recommendation under which circumstances it might be acceptable not to enable this feature?

johngian:
The reason OIDC uses nonce is to protect you from replay attacks. The idea is that the application generates the nonce and sends it to the OIDC provider on authentication request. Then the provider includes this nonce to the callback so you can correlate the authentication response to the initial request.

What we've noticed in some apps is that we get the same error when a user initiates the authentication from tab A and opens the callback to tab B on a different session (eg, different browser or container). Session in B doesn't have the nonce value and then django throws this error.

johngian:
Also yes, your assumption also sounds plausible.

cfra:
I have to admit I am not too familiar with OIDC, but from my understanding of OAuth 2.0, when control is passed back from the authentication provider to our callback, our application is passed an authorization code which it will exchange for an access token.

That exchange is only possible once, if there is another attempt to exchange the same authorization code for an access token, all access tokens previously generated for this access code will be invalidated.

Doesn't this already protect against replay attacks?
What is an attack scenario where the nonce provides better security?

johngian:
An example from the top of my head is:

* User A initiates authentication flow
* User A gets redirected to OP
* Malicious user B sniffs the `id_token` from the network
* Replays the authentication flow to the RP before user A and authenticates successfully

With `nonce` associated with user A session, user B cannot authenticate to the RP because `nonce` cannot be validated.

florian-hartmann:
> What we've noticed in some apps is that we get the same error when a user initiates the authentication from tab A and opens the callback to tab B on a different session (eg, different browser or container). Session in B doesn't have the nonce value and then django throws this error.

[...]

PR #345, opened 2020-02-04
Allow multiple parallel login sessions

Flor1an-dev:
This merge request allows multiple parallel login sessions.

This MR should fix the issues #303 and #312.

**The problem**
Currently `mozilla-django-oidc` only allows one parallel session to go through the OpenID Connect login flow. The problem is, that the `state` and `nonce` parameters are stored in two session variables.

Here is an example workflow where the user runs into an error:
1. `Tab A` User wants to log in, `state` and `nonce` are set in the user's session and user is redirected to the OIDC Provider.
2. `Tab B` User also wants to log in a second tab (of the same application), e.g. because he restores his browser session after reboot etc. This will generate a new `state` and `nonce` value and overwrite the existing two in the session.
3. `Tab A` User authenticates at the OIDC Provider and is redirected back with the `state` and `nonce` from the first step. As those were overwritten, `mozilla-django-oidc` fails with an error that the `state` does not match. `mozilla-django-oidc` also deletes the `nonce` from the session.
4. `Tab B` User also authenticates against the OIDC Provider, is redirected back with the `state` and `nonce` from step 2. The `state` matches the `state` in the session but the `nonce` was deleted in the 3. step, so the user will get an error as well.

=> The problem is that `state` and `nonce` are stored in two variables and will be overwritten when multiple tabs are used.

**Solution**
I've replaced the two session variables `oidc_state` and `oidc_nonce` with a dictionary `oidc_states`.
This dictionary contains `state` (key) - `nonce` (value) pairs.
Every time the user wants to log in, a new `state` - `nonce` combination will be added to this dictionary without overwriting other values.

When a user comes back from the OIDC Provider, `mozilla-django-oidc` will check for the `state` if there is a valid combination in the `oidc_states` dictionary. If yes it will also check the `nonce` value.

**Drawbacks**
- Previously, when the user came back, the `nonce` value was deleted immediately to avoid using it multiple times. This also happened if no `state` or a wrong `state` was provided. Now we have multiple `state` / `nonce` values, and we first need to identify the correct pair. So if no or a wrong `state` are given, we cannot delete a nonce value. But the `nonce` will be deleted if regularly used during a login and thus, I think this can be accepted from a security perspective.
- The `oidc_states` dictionary in the session can grow over time (e.g. if the user initializes a login session but never ends it). For this reason, I empty the whole dictionary if there are more than 20 `state` / `nonce` combinations. I didn't want to implement an aging mechanism to keep the code as simple as possible. Having more than 20 parallel login session is unlikely (or we can also increase this value) and I think we can also accept a login failure in case this really happens.

I adapted the Unittests and wrote an additional one which adds two `state` / `nonce` combinations (simulates two parallel login sessions), executes one successful login and then checks that the other combination is still left in the `oidc_states` dictionary.

### DIFF
--- a/mozilla_django_oidc/utils.py
+++ b/mozilla_django_oidc/utils.py
@@ -1,5 +1,5 @@
-from datetime import datetime
 import logging
+import time
 import warnings
 
 try:
@@ -84,7 +84,7 @@ def add_state_and_nonce_to_session(request, state, params):
     if len(request.session['oidc_states']) >= limit:
         LOGGER.warning('User has more than {} "oidc_states" in his session, deleting the oldest one!'.format(limit))
         oldest_state = None
-        oldest_added_on = datetime.now().timestamp()  # use float because datetime is not JSON serializable
+        oldest_added_on = time.time()
         for item_state, item in request.session['oidc_states'].items():
             if item['added_on'] < oldest_added_on:
                 oldest_state = item_state
@@ -94,5 +94,5 @@ def add_state_and_nonce_to_session(request, state, params):
 
     request.session['oidc_states'][state] = {
         'nonce': nonce,
-        'added_on': datetime.now().timestamp(),  # use float because datetime is not JSON serializable
+        'added_on': time.time(),
     }
--- a/mozilla_django_oidc/utils.py
+++ b/mozilla_django_oidc/utils.py
@@ -8,6 +8,7 @@ except ImportError:
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
+from django.utils.crypto import get_random_string
 
 
 def parse_www_authenticate_header(header):
@@ -51,3 +52,26 @@ def is_authenticated(user):
     msg = '`is_authenticated()` is going to be removed in mozilla-django-oidc v 2.x'
     warnings.warn(msg, DeprecationWarning)
     return user.is_authenticated
+
+
+def add_state_and_nonce_to_session(request, state, params):
+    nonce = None
+    if import_from_settings('OIDC_USE_NONCE', True):
+        nonce = get_random_string(import_from_settings('OIDC_NONCE_SIZE', 32))
+        params.update({
+            'nonce': nonce
+        })
+
+    # Store Nonce with the state parameter in the session "oidc_states" dictionary.
+    # The dictionary can store multiple state/nonce combinations to allow parallel
+    # authentication flows which would otherwise overwrite state/nonce values!
+
+    # Initialize 'oidc_states' dictionary. Make sure the dictionary does not get
+    # too big by resetting the dictionary if there are more than 20 entries
+    # (unlikely to reach so many parallel login sessions at the same time).
+    if 'oidc_states' not in request.session or \
+            not isinstance(request.session['oidc_states'], dict) or \
+            len(request.session['oidc_states']) > 20:
+        request.session['oidc_states'] = {}
+
+    request.session['oidc_states'][state] = nonce
--- a/mozilla_django_oidc/utils.py
+++ b/mozilla_django_oidc/utils.py
@@ -60,6 +60,12 @@ def is_authenticated(user):
 
 
 def add_state_and_nonce_to_session(request, state, params):
+    """
+    Stores the `state` and `nonce` parameters in a session dictionary including the time when it was added.
+    The dictionary can contain multiple state/nonce combinations to allow parallel logins with multiple browser
+    sessions.
+    To keep the session space to a reasonable size, the dictionary is kept at 50 state/nonce combinations maximum.
+    """
     nonce = None
     if import_from_settings('OIDC_USE_NONCE', True):
         nonce = get_random_string(import_from_settings('OIDC_NONCE_SIZE', 32))
--- a/mozilla_django_oidc/utils.py
+++ b/mozilla_django_oidc/utils.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+import logging
 import warnings
 
 try:
@@ -9,6 +11,9 @@ except ImportError:
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.crypto import get_random_string
+
+
+LOGGER = logging.getLogger(__name__)
 
 
 def parse_www_authenticate_header(header):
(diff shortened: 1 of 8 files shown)


rustanacexd:
I tested this PR, it somewhat works but still doesn't allow parallel multiple login sessions.  I am still getting `OIDC callback state not found in session `oidc_states`!`

Flor1an-dev:
@rustanacexd Can you reproduce this issue? 

I can think of 2 situations where this error could appear:
- User Session gets cleared, that means the OIDC states are lost in the session.
- There are too many OIDC state values stored in the Session (currently > 20), in that case the states are also cleaned up.

I've currently not a good idea how to deal with the later problem, as this PR stores multiple states, we should take care that the session does not get filled up, so some cleaning or limiting is necessary. Maybe it makes sense to have a FIFO mechanism for this instead of deleting all states.

I'll do some changes to the PR so that the two situations above will create two different kind of error messages to make it easier to debug the issue.

Flor1an-dev:
@rustanacexd I've done some more changes, please check them out and check if you are still getting the errors.

I've increased the number of state/nonce values in the session form 20 to 50 and made it configurable via the `OIDC_MAX_STATES` setting (default 50).
The state/nonce combinations now also store the timestamp when they were added to the session. When there are more than 50 combinations in the session, the library will delete the oldest state and keep the latest added states in the session. Hopefully this will solve most situations but in the end, there might still be situations when there are just too many states in the session and some get lost/deleted.

mozilla_django_oidc/views.py
```diff
@@ -68,18 +64,28 @@ def get(self, request):
...
+            # State is the key, the value is a dictionary with the Nonce in the "nonce" field.
+            state = request.GET.get('state')
+            if state not in request.session['oidc_states']:
+                msg = 'OIDC callback state not found in session `oidc_states`!'
                 raise SuspiciousOperation(msg)
 
+            # Get the nonce from the dictionary for further processing and delete the entry to
+            # prevent replay attacks.
+            nonce = request.session['oidc_states'][state]['nonce']
+            del request.session['oidc_states'][state]
```
istreeter: Is it possible to move this `del` much higher up in the method?  In particular, before you you test if `error` or `code` are in the query parameters.

If the purpose of a nonce is to prevent replay attacks, then it feels safer to remove the nonce from the `oidc_states` as early as you possibly can.  Right now I cannot figure out if an attacker could take advantage of your implementation (probably not!).  But if you can move it further up then you might prevent bugs or replay attack vulnerabilities from creeping into the code in future.

Flor1an-dev: I don't think we can move it further up. All the code above does, is to retrieve the state/nonce combination from the dictionary.

I'm actually not really sure how much the nonce helps here anyways.

RubenGarcia:
I would also be interested in this patch being accepted.

akatsoulas:
Thanks @Flor1an-dev for this PR. This should be reviewed by the end of this week.

mozilla_django_oidc/utils.py
```diff
@@ -51,3 +57,46 @@ def is_authenticated(user):
     msg = '`is_authenticated()` is going to be removed in mozilla-django-oidc v 2.x'
     warnings.warn(msg, DeprecationWarning)
     return user.is_authenticated
+
+
+def add_state_and_nonce_to_session(request, state, params):
```
akatsoulas, line 62: Could you add a docstring in this method or move the comments below here as a docstring?

mozilla_django_oidc/utils.py
```diff
@@ -51,3 +57,46 @@ def is_authenticated(user):
...
+    nonce = None
+    if import_from_settings('OIDC_USE_NONCE', True):
```
akatsoulas, line 64: It would be better to leverage the staticmethod `get_settings` which can be used to configure the OIDC lib for multiple providers.

mozilla_django_oidc/utils.py
```diff
@@ -51,3 +57,46 @@ def is_authenticated(user):
...
+    # With this value, the oldest element can be found and deleted from the session.
+    if 'oidc_states' not in request.session or \
+            not isinstance(request.session['oidc_states'], dict):
+        request.session['oidc_states'] = {}
+
+    # Make sure that the State/Nonce dictionary in the session does not get too big.
+    # If the number of State/Nonce combinations reaches a certain threshold, remove the oldest
+    # state by finding out
+    # which element has the oldest "add_on" time.
+    limit = import_from_settings('OIDC_MAX_STATES', 50)
```
akatsoulas, line 84: Similar to the comment above, this should use the `get_settings`.

istreeter:
@akatsoulas My company is already running this code in production **However** I had to patch it first.  Please see [this commit](https://github.com/DouglasConnect/mozilla-django-oidc/commit/66b72e74aab0af0051661c80e04c6bbffd68eeac) for the patch I had to apply.

Without my patch, this PR can generate SuspiciousOperation exceptions, for example when the user has multiple browser tabs open at the same time.  There is a race condition when multiple tabs try to authenticate at the same time.

jwhitlock:
@Flor1an-dev are you interested in responding to the feedback? If not, I'd start a new PR to make the changes to get this through code review.

mozilla_django_oidc/utils.py
Flor1an-dev: Done.

Flor1an-dev:
> @akatsoulas My company is already running this code in production **However** I had to patch it first. Please see [this commit](https://github.com/DouglasConnect/mozilla-django-oidc/commit/35187ff48cf2b9d6f496444e792e931e2c51c1da) for the patch I had to apply.
> 
> Without my patch, this PR can generate SuspiciousOperation exceptions, for example when the user has multiple browser tabs open at the same time. There is a race condition when multiple tabs try to authenticate at the same time.

Good catch, I've integrated your commit! Thx.

mozilla_django_oidc/utils.py
Flor1an-dev: @akatsoulas: There is no `get_settings` method in the `utils.py` file. And when I would use one of the existing implemented `get_settings` methods, we would have a circular dependency.
Any suggestions?

Flor1an-dev: I used the method like in `views.py` `get_next_url` or in `urls.py`.

akatsoulas:
I am closing this in favor of #358 which is the same PR with the commits merged. Thanks @Flor1an-dev and @istreeter for working on this!

PR #350, opened 2020-04-15
Fix JWKS handling with same `kid` but different `alg` across JWKs

davidjb:
Prior to this change, when a JWKS was delivered with multiple JWK with the same `kid` but different `alg` specified, the `retrieve_matching_jwk` function would only consider the first key in the list where the `kid` matched.  If the `alg` didn't match, an exception would be raised and if there was a valid JWK with the right `alg` (eg later in the list) it would not be considered.

In the JWK spec (https://tools.ietf.org/html/rfc7517#section-4.4), the `alg` value isn't required to be unique -- only to match -- so the check shouldn't raise an exception at this point, but rather continue checking all remaining JWKs.

With this change, an exception is still raised at the end if no valid JWK is found.  Test output was updated and a new test added to ensure the right key is retrieved from a list with multiple matching `kids` (`kid` need not be unique as per [rfc7517](https://tools.ietf.org/html/rfc7517#section-4.5)).

This follows and improves the change at #256 and issue at #247.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,12 @@
 History
 -------
 
+1.2.4 (Unreleased)
+==================
+
+* Fix JWKS handling when the same `kid` value is used across JWKs with
+  different `alg` specified
+
 1.2.3 (2020-01-02)
 ===================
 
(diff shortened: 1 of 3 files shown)


davidjb:
Any thoughts on being able to merge this in?

PR #357, opened 2020-07-16
Allow multiple parallel login sessions

akatsoulas:


akatsoulas:
@johngian @Flor1an-dev @jwhitlock could you take a look at this? Basically I moved the nonce configuration outside of the new method added by Florian.

mozilla_django_oidc/utils.py
```diff
@@ -67,12 +67,7 @@ def add_state_and_nonce_to_session(request, state, params):
     To keep the session space to a reasonable size, the dictionary is kept at 50 state/nonce
     combinations maximum.
     """
-    nonce = None
-    if import_from_settings('OIDC_USE_NONCE', True):
-        nonce = get_random_string(import_from_settings('OIDC_NONCE_SIZE', 32))
```
jwhitlock: `get_random_string` is now unused, so the import should be removed.

PR #358, opened 2020-07-16
Support multiple sessions

akatsoulas:


akatsoulas:
I opened this PR after discussing a bit through slack with @jwhitlock. It's Florian's work squashed in a single commit and the suggestion from Ian. If there aren't comments, this can be directly merged to master. @jwhitlock It would be great if you had some time for a final pass. I approved #345 on my end

mozilla_django_oidc/views.py
```diff
@@ -14,7 +14,8 @@
 from django.utils.module_loading import import_string
 from django.views.generic import View
 
-from mozilla_django_oidc.utils import absolutify, import_from_settings
+from mozilla_django_oidc.utils import absolutify, import_from_settings,\
+    add_state_and_nonce_to_session
```
jwhitlock, line 18: Nit: I prefer parenthesis again

mozilla_django_oidc/utils.py
```diff
@@ -51,3 +57,54 @@ def is_authenticated(user):
...
+        request.session['oidc_states'] = {}
+
+    # Make sure that the State/Nonce dictionary in the session does not get too big.
+    # If the number of State/Nonce combinations reaches a certain threshold, remove the oldest
+    # state by finding out
+    # which element has the oldest "add_on" time.
+    limit = import_from_settings('OIDC_MAX_OIDC_STATES', 50)
+    limit = import_from_settings('OIDC_MAX_STATES', 50)
+    if len(request.session['oidc_states']) >= limit:
+        LOGGER.warning(
```
jwhitlock, line 94: Nit: I think this should be an `.INFO` level log, but maybe we don't expect it in production, only if `OIDC_MAX_STATES` is too low?

mozilla_django_oidc/views.py
```diff
@@ -68,18 +64,36 @@ def get(self, request):
...
+            # State is the key, the value is a dictionary with the Nonce in the "nonce" field.
+            state = request.GET.get('state')
+            if state not in request.session['oidc_states']:
+                msg = 'OIDC callback state not found in session `oidc_states`!'
                 raise SuspiciousOperation(msg)
 
+            # Get the nonce from the dictionary for further processing and delete the entry to
+            # prevent replay attacks.
+            nonce = request.session['oidc_states'][state]['nonce']
+            del request.session['oidc_states'][state]
```
jwhitlock, line 82: 👍 good job cleaning up nonce on usage

tests/test_utils.py
```diff
@@ -1,8 +1,10 @@
+from django.contrib.sessions.middleware import SessionMiddleware
 from django.core.exceptions import ImproperlyConfigured
 from django.test import TestCase, override_settings
 from django.test.client import RequestFactory
 
-from mozilla_django_oidc.utils import absolutify, import_from_settings
+from mozilla_django_oidc.utils import absolutify, add_state_and_nonce_to_session, \
```
jwhitlock, line 6: Nit: one more place I would use parenthesis instead of line continuation.

mozilla_django_oidc/utils.py
```diff
@@ -51,3 +57,54 @@ def is_authenticated(user):
...
+    # With this value, the oldest element can be found and deleted from the session.
+    if 'oidc_states' not in request.session or \
+            not isinstance(request.session['oidc_states'], dict):
+        request.session['oidc_states'] = {}
+
+    # Make sure that the State/Nonce dictionary in the session does not get too big.
+    # If the number of State/Nonce combinations reaches a certain threshold, remove the oldest
+    # state by finding out
+    # which element has the oldest "add_on" time.
+    limit = import_from_settings('OIDC_MAX_OIDC_STATES', 50)
```
jwhitlock, line 91: This line should be removed, it looks like an old version of the next line.

tests/test_utils.py
```diff
@@ -47,3 +49,102 @@ def test_absolutify_path_host_injection(self):
         ).get('/', SERVER_PORT=443)
         url = absolutify(req, 'evil.com/foo/bar')
         self.assertEqual(url, 'https://testserver/evil.com/foo/bar')
+
+
+class SessionStateTestCase(TestCase):
```
jwhitlock, line 54: 👍 I like this set of tests, good exercise of `add_state_and_nonce_to_session`

tests/test_views.py
```diff
@@ -37,8 +39,10 @@ def test_get_auth_success(self):
         }
         url = reverse('oidc_authentication_callback')
         request = self.factory.get(url, get_data)
-        request.session = {
-            'oidc_state': 'example_state'
+        client = Client()
```
jwhitlock, line 42: I'm curious why it was necessary to copy a session from ``Client()``. I have wrestled with sessions in a test environment however, so I'm not too surprised.

tests/test_middleware.py
```diff
@@ -57,8 +57,10 @@ def test_is_POST(self):
     @override_settings(OIDC_RP_CLIENT_ID='foo')
     @override_settings(OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS=120)
     @patch('mozilla_django_oidc.middleware.get_random_string')
-    def test_is_ajax(self, mock_random_string):
-        mock_random_string.return_value = 'examplestring'
+    @patch('mozilla_django_oidc.utils.get_random_string')
+    def test_is_ajax(self, mock_utils_random, mock_middleware_random):
+        mock_middleware_random.return_value = 'examplestring'
```
jwhitlock, line 62: For these and others, I suspect the mock of ``mozilla_django_oidc.middleware.get_random_string`` is no longer needed.

akatsoulas: I believe this is used to mock the `state` attribute in the views. Before this PR both `state` and `nonce` had the same value in tests. I will update the string in a following PR to make this clearer.

tests/test_views.py
Flor1an-dev: The requests object created by the factory does not have a session attribute! I think there are two possible ways to get a session attribute:
1. Like above, use the existing session object from the client. As the `Client` class is based on `RequestFactory`, we could use a `Client` directory instead of the `RequestFactory`.

2. We could do it similar like here: https://github.com/akatsoulas/mozilla-django-oidc/blob/b18fbd00240f89f02528363da2af61f881e28d26/tests/test_utils.py#L60
But this would require executing the session in each test method as the URL/data parameters do change in every test method compared to the test_utils.py where the request is always the same.

jwhitlock: That sounds familiar, and the [Django docs](https://docs.djangoproject.com/en/3.0/topics/testing/advanced/#django.test.RequestFactory) confirm it. A little weirdness is expected when testing authentication code.

tests/test_middleware.py
jwhitlock: No, you are right, I see where `examplestring` is being asserted later in the test, it still needs to be mocked. I was fooled by GitHub's context diff again. But, if we had a time machine, `examplestate` or `fakestate` would be better.

PR #359, opened 2020-07-17
Move nonce outside of add_state_and_noce_to_session method.

akatsoulas:
This is a follow up on #345 and #358 . @jwhitlock / @johngian quick r?

### DIFF
--- a/tests/test_middleware.py
+++ b/tests/test_middleware.py
@@ -57,10 +57,8 @@ class SessionRefreshTokenMiddlewareTestCase(TestCase):
     @override_settings(OIDC_RP_CLIENT_ID='foo')
     @override_settings(OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS=120)
     @patch('mozilla_django_oidc.middleware.get_random_string')
-    @patch('mozilla_django_oidc.utils.get_random_string')
-    def test_is_ajax(self, mock_utils_random, mock_middleware_random):
+    def test_is_ajax(self, mock_middleware_random):
         mock_middleware_random.return_value = 'examplestring'
-        mock_utils_random.return_value = 'examplenonce'
 
         request = self.factory.get(
             '/foo',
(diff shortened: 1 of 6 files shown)


tests/test_middleware.py
```diff
@@ -57,10 +57,8 @@ def test_is_POST(self):
     @override_settings(OIDC_RP_CLIENT_ID='foo')
     @override_settings(OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS=120)
     @patch('mozilla_django_oidc.middleware.get_random_string')
-    @patch('mozilla_django_oidc.utils.get_random_string')
-    def test_is_ajax(self, mock_utils_random, mock_middleware_random):
+    def test_is_ajax(self, mock_middleware_random):
         mock_middleware_random.return_value = 'examplestring'
```
jwhitlock, line 61: Nit: use `side_effect` to get two different strings:

```suggestion
        mock_middleware_random.side_effect = ['examplenonce', 'examplestring']
```

akatsoulas:
> The code looks good, but could you explain more about why you are moving the code out of `add_state_and_nonce_to_session`? It duplicates the nonce code and makes it more likely that the methods will diverge in the future.

Indeed this is not ideal at the moment but this way the nonce creation can use the `get_settings` static method. With this method, you can configure the OIDC lib for two different OIDC providers. This won't be a problem if you have similar settings but it's not going to work in the case where the RP is configured with `nonce` for provider A and without for provider B. 

That said, I am open to any ideas that can bundle this code into a single place without complicating the code base too much.

jwhitlock:
Thanks for the explanation, I completely missed the change from ``import_from_settings`` to ``get_settings``. This makes more sense now.

One way to unify the code would be to pass the output if the ``get_settings`` call to a common function. However, that seems more awkward than useful. Another way would be to move the function to a base class or mixin, but I find that also is awkward, and can make it harder to understand the code.

I've used a pair of comment in the past, like ``keep this logic in sync with SessionRefresh.process_request``, as an aid to future devs (maybe me). It is easier to write that comment if the code that needs to stay in sync is in a small function, like ``get_optional_nonce``. But again, that may be overkill for this case.

Or, it can be good to keep awkward stuff around, suggesting ways that the code should be re-arranged to make stuff less awkward in the future, leading to a big refactor that seems obvious in hindsight.

That is a long way of saying 👍, thanks for the explanation, no changes needed at this time.

tests/test_middleware.py
akatsoulas: This has multiple occurrences so I will follow up with a different PR to change all of them

PR #364, opened 2020-07-31
Update middleware.py to not set oidc_login_next for AJAX 403 redirects

drwonky:
Moved the oidc_login_next to below the AJAX return, this means AJAX requests will be redirected to the LOGIN_REDIRECT_URL instead of the API endpoint that generated the session refresh.

Current behavior is to redirect AJAX 403 responses to the API endpoint, this results in the API endpoint response being displayed in the browser main window.

As noted, redirects with XHR requests are extremely finicky, in most cases a simple 302 redirect results in a CORS denial.  So the only way to get a session refresh is to do a window.location.href redirect, the 403 becomes a "simple" 302 redirect (using CORS terminology for simple vs XHR requests).

### DIFF
--- a/mozilla_django_oidc/middleware.py
+++ b/mozilla_django_oidc/middleware.py
@@ -88,7 +88,7 @@ class SessionRefresh(MiddlewareMixin):
 
         now = time.time()
 
-        if 'X-Refresh-OIDC-Token' in request.headers:
+        if hasattr(request,'headers') and 'X-Refresh-OIDC-Token' in request.headers:
             request.session['oidc_id_token_expiration'] = now
 
         expiration = request.session.get('oidc_id_token_expiration', 0)
--- a/mozilla_django_oidc/middleware.py
+++ b/mozilla_django_oidc/middleware.py
@@ -86,8 +86,13 @@ class SessionRefresh(MiddlewareMixin):
             LOGGER.debug('request is not refreshable')
             return
 
-        expiration = request.session.get('oidc_id_token_expiration', 0)
         now = time.time()
+
+        if 'X-Refresh-OIDC-Token' in request.headers:
+            request.session['oidc_id_token_expiration'] = now
+
+        expiration = request.session.get('oidc_id_token_expiration', 0)
+
         if expiration > now:
             # The id_token is still valid, so we don't have to do anything.
             LOGGER.debug('id token is still valid (%s > %s)', expiration, now)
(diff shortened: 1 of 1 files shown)


akatsoulas:
@davidjb thanks for this PR. Could you please update the commit message to something more descriptive?

davidjb:
@akatsoulas It's not my PR, but I'd definitely appreciate a review of mine over at #350 😄

akatsoulas:
Yep you are right :) Autocomplete mix up. I will have a look at the PR

akatsoulas:
@drwonky Could you update the commit message to something more descriptive?

drwonky:
Hi, I amended the commit and changed the message, please let me know if it is acceptable.  Sorry for the commit spam, but I made the original change online through github, so I had to clone it offline and amend it there.

drwonky:
Hi, sorry about the way this pull request is going, it's a bit of a mess because I'm getting used to the way Github does things.

If you want to cherry pick my 403 commit, that's fine.

My latest commit is a really simple hack to the middleware that allows AJAX clients to proactively refresh session tokens.

The client merely needs to add the header X-Refresh-OIDC-Token to the request and it will cause the middleware to reset the oidc_id_token_expiration to now, which will trigger the token reauth logic.

This allows SPAs and Javascript based applications to control when they will be interrupted for a session refresh, so the control flow won't be surrendered to an inopportune redirect.

Some might see this as a hack, and it is, but it's also an elegant and simple solution to a forced session refresh.  Now you don't have to force a logout to refresh your token.  The middleware does what it is supposed to, but now the behavior can be influenced by the client.

The value of the header is inconsequential, merely the presence of the header is required.

If you want a use case, let's say when you authenticate for the first time, the browser knows how long the session will last, then it periodically checks the remaining time and informs the user the session token will timeout at a specific time.  The app can then allow the user to continue working or perform a session token refresh then, avoiding interruption.

As with the 403 patch, this is no different, you get redirected to the LOGIN_REDIRECT_URL.

drwonky:
Hi there, since this PR seems to have stalled, would you like me to close this and redo the edits as branches and open 2 new pull requests?  I don't know if you want my second pull request, since that implementation might be not be desired in your grand design.

akatsoulas:
@drwonky yes that would be the ideal scenario. Let's get two PRs and we can discuss the second fix separately from the `oidc_login_next`. Thank you for this patch!

PR #368, opened 2020-08-07
Update History for 1.2.4 release.

akatsoulas:


HISTORY.rst
```diff
@@ -3,14 +3,23 @@
 History
 -------
 
-1.2.4 (Unreleased)
+1.2.4 (update-with-a-date)
 ==================
 
+* Fix error in README.rst
+  Thanks `@der-gabe`_
```
jwhitlock, line 10: This format requires a latter line that says what the link is, like:

```
.. _`@der-gabe`: https://github.com/der-gabe
```

The primary benefit is that the prose isn't littered with hyperlinks and weird line breaks, and you can link to the same word twice. I think that's why I used it, but maybe it is more confusing now and in the future.

Alternatively, you can use inline links, like:
```suggestion
  Thanks `@der-gabe <https://github.com/der-gabe>`_
```

This might be better for the history doc, since releases are rare and "Thanks `@foo <https://github.com/foo>`" will easily fit on a line.

Also, no matter how much I work with RST / Sphinx, I always have to go find documentation on the syntax format. It just doesn't stick in my memory like Markdown. But, it has more features than MD for docs, so it is still worth struggling with.

PR #374, opened 2020-08-31
Fix encoding of scopes 

sannies:


akatsoulas:
@johngian could you take a look at this?

johngian:
From docs:
```
 urllib.parse.urlencode(query, doseq=False, safe='', encoding=None, errors=None, quote_via=quote_plus)
```

I think instead of manually re-implementing the functionality you could pass the `quote` function as `quote_via=quote`.

johngian:
Here is a request example from the OIDC spec docs:

```
  HTTP/1.1 302 Found
  Location: https://server.example.com/authorize?
    response_type=code
    &scope=openid%20profile%20email
    &client_id=s6BhdRkqt3
    &state=af0ifjsldkj
    &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb
```

I guess it makes sense to use escaped spaces instead of plus signs.

johngian:
Thanks @sannies for the contribution. Closing this one after #404

PR #377, opened 2020-10-05
Add RefreshOIDCAccessToken middleware

diurnalist:
**Note**: this was originally proposed as #301. I took the nice work by @GermanoGuerrini and compressed it into a single commit, removed the HISTORY.rst change, and also changed one thing: now, the id_token is dropped from the session, because it cannot be verified securely. This is called out in the documentation. In my use-case, and perhaps many others, this is not a problem, as the access_token is the most important thing to preserve, as it's used to authenticate on behalf of the user to other services.

---

The OP can provide a refresh_token to the client on authentication. This
can later be used to get a new access_token. Typically refresh_tokens
have a longer TTL than access_tokens and represent the total allowed
session length. As a bonus, the refresh happens in the background and
does not require taking the user to a new location (which also makes it
more compatible with e.g., XHR).

This adds a new middleware to perform the refreshing of the access token
in the background.


diurnalist:
@akatsoulas , I think the other PR had been sitting for quite some time, if you get a chance to review this (and maybe @GermanoGuerrini as well?), maybe it can move forward this time.

Of course, feel free to close this and address #301 instead. I just wanted this functionality for my own deploy, saw it was already done (woo!) and then took the liberty of cleaning it up in hopes that it could make a merge soon.

👋 🚪 🚶‍♂️

diurnalist:
I've also added error handling for when the refresh fails... prior, it would bomb the entire request, which to me is not expected.

tevariou:
Any update on that?

mozilla_django_oidc/middleware.py
```diff
@@ -179,3 +189,59 @@ def process_request(self, request):
...
+            )
+            response.raise_for_status()
+            token_info = response.json()
+        except requests.exceptions.Timeout:
+            LOGGER.debug('timed out refreshing access token')
+            return
+        except requests.exceptions.HTTPError as exc:
+            LOGGER.debug('http error %s when refreshing access token',
+                exc.response.status_code)
+            return
```
tevariou, line 232: 1 - Maybe the user should be logged out at this point if the status_code == 401
2 - An appropriate response should be sent in case the request is an ajax call

mozilla_django_oidc/middleware.py
```diff
@@ -179,3 +189,59 @@ def process_request(self, request):
...
+
+        token_payload = {
+            'grant_type': 'refresh_token',
+            'client_id': client_id,
+            'client_secret': client_secret,
+            'refresh_token': refresh_token,
+        }
+
+        try:
+            response = requests.post(
```
tevariou, line 219: It doesn't handle the case when OIDC_TOKEN_USE_BASIC_AUTH is True

tevariou (changes requested):

julwin (approved):
lgtm!

julwin:
Can we have a look at this PR? This is a much needed feature..

oar-spease:
What is needed for this PR to be looked at? It's kind of a crucial piece of any integration of OIDC into django unless you like constantly logging in over and over, and it's just been sitting here. 

It would be nice to get this merged so I can stop maintaining my own branch- what exactly needs done here?

diurnalist:
For those interested in (or even using) this patch, I pushed a fix for an issue we found, where it was not prompting for reauth if the refresh token request returned a 400 error. I had assumed incorrectly that a 401 would be returned, but that is not in the OAuth 2.0 spec.

MrAlexBailey:
> lgtm!

+1, what can we help with maintainers?

GermanoGuerrini (approved):
You did an excellent job. I'd like to see it merged too.

mozilla_django_oidc/middleware.py
```diff
@@ -152,26 +204,85 @@ def process_request(self, request):
...
+                'unknown error occurred when refreshing access token: %s', exc)
+            # Don't prompt for reauth as this could be a temporary problem
+            return self.finish(request, prompt_reauth=False)
+
+        # Until we can properly validate an ID token on the refresh response
+        # per the spec[1], we intentionally drop the id_token.
+        # [1]: https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse
+        id_token = None
+        access_token = token_info.get('access_token')
+        refresh_token = token_info.get('refresh_token')
```
AndrejZbin, line 287: The refresh token is optional. 
https://www.ietf.org/rfc/rfc6749.txt

So this doesn't work, for example, for amazon cognito, as they don't return new refresh token.
https://docs.aws.amazon.com/cognito/latest/developerguide/token-endpoint.html

I think thee refresh_token should only be stored when it's contained in the response, otherwise keep the old one.

AndrejZbin (changes requested):
Great job with this, it's very useful to us. We would like to see this merged.

I noticed one issue with the implementation when using this for Amazon Cognito.

mozilla_django_oidc/views.py
```diff
@@ -49,9 +49,13 @@ def login_success(self):
...
-        # using the RenewIDToken middleware.
-        expiration_interval = self.get_settings('OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS', 60 * 15)
-        self.request.session['oidc_id_token_expiration'] = time.time() + expiration_interval
+        # using the SessionRefresh or RefreshOIDCAccessToken middlewares.
+        expiration_interval = self.get_settings(
+            'OIDC_RENEW_TOKEN_EXPIRY_SECONDS',
+            # Handle old configuration value
+            self.get_settings('OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS', 60 * 15)
+        )
+        self.request.session['oidc_token_expiration'] = time.time() + expiration_interval
```
tevariou, line 58:
```suggestion
        self.request.session['oidc_id_token_expiration'] = time.time() + expiration_interval
```

tevariou (changes requested):

nt-unity:
Any updates on this please? Would really like to see this added, it is essential.

moyoakala (approved):
Looks good to me

dab-ib:
Are there any updates on this? The whole topic is very interesting and could solve many of my XHR problems without writing custom code.

tests/test_middleware.py
```diff
@@ -133,6 +133,65 @@ def test_expired_token_forces_renewal(self, mock_middleware_random):
         self.assertEqual(expected_query, parse_qs(qs))
 
 
+@override_settings(OIDC_OP_AUTHORIZATION_ENDPOINT='http://example.com/authorize')
+@override_settings(OIDC_RP_CLIENT_ID='foo')
+@override_settings(OIDC_RENEW_TOKEN_EXPIRY_SECONDS=120)
+class RefreshOIDCAccessTokenMiddlewareTestCase(TestCase):
+    def setUp(self):
+        self.factory = RequestFactory()
+        self.middleware = RefreshOIDCAccessToken()
```
phoebem, line 142: RefreshOIDCAccessToken() request an arg, should be something like:
  self.middleware = RefreshOIDCAccessToken(MagicMock)

mozilla_django_oidc/middleware.py
```diff
@@ -152,26 +204,85 @@ def process_request(self, request):
...
+            return self.finish(request, prompt_reauth=False)
+        except Exception as exc:
+            LOGGER.debug(
+                'unknown error occurred when refreshing access token: %s', exc)
+            # Don't prompt for reauth as this could be a temporary problem
+            return self.finish(request, prompt_reauth=False)
+
+        # Until we can properly validate an ID token on the refresh response
+        # per the spec[1], we intentionally drop the id_token.
+        # [1]: https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse
```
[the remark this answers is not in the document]
muety: In my understanding, what we could do is first verify the token's integrity analogously to [here](https://github.com/mozilla/mozilla-django-oidc/blob/774b140b9311c6c874c199bfdb266e51f36740a7/mozilla_django_oidc/auth.py#L192) and then additionally check `iss`, `sub`, `iat` and `aud` in comparison with the old token (available via `request.session.get('oidc_id_token')` if `OIDC_STORE_ID_TOKEN=True`).

Issue #348, opened 2020-04-06
Django CSRF token invalid after session refresh


bsteverink:
Please point me in the right direction if this is not the proper place to post this issue. 

I have implemented mozilla-django-oidc 1.2.3 in several of my projects. However when the user takes more time that the session expiration time to fill in a form, the csrf token that has been added to the form at page load time is invalidated (a new token is created by Django on session refresh if I understand correctly). 

Should the mozilla-django-oidc library provide a mechanism to keep the csrf token in sync (would involve some serious hacking I would say), or am I missing something?

Any pointers/help is appreciated.

drwonky:
The csrf token is not generated or manipulated by mozilla_django_oidc.  What's happening is that you have the middleware enabled and the session is being reauthed after 15 minutes.  The middleware only runs on GET method requests, so a form submission via POST should still be valid.  I suggest you record the network interaction in your browser, because I suspect your problem isn't the middleware if POST is affected.

crgwbr:
@drwonky @bsteverink I've seen the same issue in my projects. For me at least, it happens as a result of multiple-tabs being open. 

1. Tab A: Load Form with CSRF token embedded in a form field
2. Tab B: Load some page
3. Wait a few minutes.
4. Tab B: Reload page. This triggers the session refresh. The `csrftoken` cookie is changed in the response from the `OIDCAuthenticationCallbackView` view.
5. Tab A: Submit form. Get _Forbidden (403) CSRF verification failed_ error.

I'm still working on figuring out why exactly `OIDCAuthenticationCallbackView` cycles the CSRF token cookie. Regardless, it seems like that's the problem we need a solution form.

bsteverink:
@crgwbr I think you are spot on in the update. This would solve my issue. The django login method generates a new CSRF token invalidating the one sent with the original POST request.

PR #378, opened 2020-10-06
Refs #348 Only call auth.login() when the user has changed

crgwbr:
Adds login to prevent calling the auth.login function when the user hasn't
actually changed. This is the case when the session is being just being
refreshed, versus a real full login. This prevents cycling the CSRF token for
the session, which thereby fixes the invalid CSRF token errors described in
bug #348.


bsteverink:
Exactly what I was looking for in #348. Can we make this move forward?

TabishB:
Would be nice to get this merged in, and ideally a package release happening as well.

daniel5gh:
Running into the same problem of CSRF token being cycled. This change looks like the correct fix; from the perspective of django this is not actually a new login, so calling `auth.login` is not required in the case of oidc session refresh. 

Is there anything required to get this reviewed? Anything I can help with?

akatsoulas (approved):
Thanks @crgwbr. Apologies this took so long.

PR #379, opened 2020-10-12
Verify user_can_authenticate when authenticating user

cfra:
The standard Django `ModelBackend` only returns a user from  `authenticate`
if `user_can_authenticate` is `True`.

The `OIDCAuthenticationBackend` doesn't check `user_can_authenticate`,
in its `authenticate` method, allowing a user to log-in that has `is_active`
set to `False`.

This behavior is a surprising deviation from the way `ModelBackend`
works.

It can lead to security problems because other important
authentication checks may inadvertently be bypassed, if the user
implements them in `user_can_authenticate`, expecting
`OIDCAuthenticationBackend` to call it the same as any other
`ModelBackend`.

It seems proper for the `OIDCAuthenticationBackend` to verify
`user_can_authenticate` and it seems to me like there are no problems
associated with doing so, so let's just call `user_can_authenticate`
to resolve the problems described above.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,12 @@ History
 pending
 =======
 
+Backwards-incompatible changes:
+
+* ``OIDCAuthenticationBackend.authenticate`` now calls ``user_can_authenticate``.
+
+Minor:
+
 * Make `get_or_create_user` compatible with custom scope configuration
   by moving scope specific code to `describe_user_by_claims`
 
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -319,7 +319,8 @@ class OIDCAuthenticationBackend(ModelBackend):
         users = self.filter_users_by_claims(user_info)
 
         if len(users) == 1:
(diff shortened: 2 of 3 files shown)


cfra:
Is there any update on this, by any chance? :thinking:

cfra:
Any chance this can be reviewed and/or discussed? It is sitting without review for over half a year now. :disappointed_relieved:

indigane:
Noticed this as well.

This could be a result of old Django conventions, but it feels dangerous to leave it up to the view since nowadays Django handles it in the authbackend, and like noted above developers may expect that of other authbackends that inherit from ``ModelBackend``.

As for the PR, is there a specific reason why the check is made in ``get_or_create_user`` and not in ``authenticate`` before calling ``get_or_create_user``?

Is the ``is_active`` check in the view necessary after this change? If, as noted above, someone has a custom ``user_can_authenticate`` then they might not want ``is_active`` checked on top of calling ``user_can_authenticate``.

Would creating an issue help get this PR forward?

PR #385, opened 2020-11-19
Users filter in docs should return collection

sauerburger:
The method `filter_users_by_claim()` should return an object that supports `len()`. The previous example in the documentation returned the user object directly. Wrapping the user in a list fixes the issue.

This is not an issues with the actual code. This PR fixes the example on how to override the user matching in the documentation.


akatsoulas (approved):

akatsoulas:
Thanks @sauerburger!

Issue #375, opened 2020-09-10
Bug: JWK `kid` attribute may not be present in the JWKS


mkotsur:
When using the RS256 algorithm with JWKS endpoint, the [plugin assumes](https://github.com/mozilla/mozilla-django-oidc/blob/14b16df291d43f175417ec107e93f5272652f687/mozilla_django_oidc/auth.py#L162) that the `kid` attribute will always be present, which is sometimes now the case. If it's not present, the code processing fails with the error below.

![image](https://user-images.githubusercontent.com/164650/92695217-712a1680-f348-11ea-871d-e0461d7a9be2.png)

While IMO it would be logical to have the attribute always present, some implementations don't have it and it's not guaranteed by the spec either:

https://tools.ietf.org/html/rfc7517#section-4.5

JWKS example:

```
{
   "keys":[
      {
         "kty":"RSA",
         "use":"sig",
         "alg":"RS256",
         "e":"AQAB",
         "n":"2dTXc9-BbJB3iR2MeAG5QNFRkJFQPyoaNK1FTMGKXdcuNCcZgR7G57VvHCC9kdv_heArKQFG_h7t35vVgtoynyIVc_WvgopEp3C1evHnuXI1oO-b6BFaQZd6yK1ySZlA2unKpSSe9x6nvEX21tcOOucrn3R_K2qW7teR2KAas3y0MJIJELWA4quS02jfCE_2MFRleVYyl-m0mP3Nlr9QjzcdqKS0nJqeh8SQjxhJTrmo8uXVig1HkTVNw4MI0rC_Z0DrPdeJJvQIR9MMKATBkLIOOu1bJvEixf3FCPKMYEw4fc5a-S7EyQ6TOyS3izszaV7Ydq_A_yAwk_9FrW4xww"
      }
   ]
}
```

A workaround is to hardcode the key details in the Django settings file, then the plugin won't try to find the key and won't stumble upon the problematic code assumption:

```
# settings.py:

OIDC_RP_IDP_SIGN_KEY = {
         "kty":"RSA",
         "use":"sig",
         "alg":"RS256",
[...]

PR #387, opened 2020-12-02
Do not fail if JWK doesn't include key id

cfra:
Section 4.5 of RFC 7517 states: "Use of [kid] is OPTIONAL".

The current code assumes that `kid` is always provided.

This is not always the case, for example the OIDC implementation of
SATOSA does not provide `kid`.

This case should be handled gracefully instead of causing a
`KeyErrror`.

Fixes: #375


akatsoulas (approved):
Thanks @cfra!

PR #394, opened 2021-02-01
Add github actions

johngian:


johngian:
@akatsoulas here is my WIP for github actions

* To follow the same paradigm with travis i am using `tox-gh-actions`. In case we don't want this as a dependency we need to work with the matrix that is offered by gh actions.
* Its only the testing part. Travis also handles the releases. But i don't have access to credentials.

akatsoulas (approved):

Issue #373, opened 2020-08-31
scopes are not encoded correctly


sannies:
Current code redirects to provider/login?scopes=scope1+scope2+scope3 but that is not correct (most providers understand it nonetheless). AWS Cognito doesn't support it and requires cognitohost/login?scopes=scope1%20scope2%20scope3.

sannies:
https://github.com/mozilla/mozilla-django-oidc/pull/374

akatsoulas:
Fixed in #404. Thanks @sannies for reporting this

PR #404, opened 2021-03-21
Fix url encoding using escaped space characters

johngian:
* urllib.parse.urlencode quotes strings using `quote_plus`
* this patch uses urllib.parse.quote instead for compatibility with some providers


johngian:
Fixes #373

johngian:
@akatsoulas r?

akatsoulas (approved):

PR #413, opened 2021-04-07
Added tests for Django 3.2

jannh:


jaap3:
Good to known the tests pass with Django 3.2 as I'm currently in the process of upgrading a project that uses this package to Django 3.2 👍

akatsoulas (approved):
thanks @jannh !

PR #420, opened 2021-06-12
Handle case that error_description is not included in www-authenticate header

adamj9431:
In the case that an OIDC provider returns 401, a www-authenticate header is included. The code breaks if the error_description field is not present in the www-authenticate header, but [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-3) says:
>In addition, the resource server MAY include the "error_description" attribute to provide developers a human-readable explanation that is not meant to be displayed to end-users.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,7 +8,8 @@ pending
 
 * Make `get_or_create_user` compatible with custom scope configuration
   by moving scope specific code to `describe_user_by_claims`
-
+* Fix bug in contrib.drf.OIDCAuthentication where `error_description` is assumed to
+  be present in `www-authenticate` header despite it being optional in the spec
 1.2.4 (2020-08-19)
 ==================
 
--- a/mozilla_django_oidc/contrib/drf.py
+++ b/mozilla_django_oidc/contrib/drf.py
@@ -79,7 +79,7 @@ class OIDCAuthentication(authentication.BaseAuthentication):
             # we can get from the www-authentication header) in the response.
             if resp.status_code == 401 and 'www-authenticate' in resp.headers:
                 data = parse_www_authenticate_header(resp.headers['www-authenticate'])
-                raise exceptions.AuthenticationFailed(data['error_description'])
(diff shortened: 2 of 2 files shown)


mozilla_django_oidc/contrib/drf.py
```diff
@@ -79,7 +79,7 @@ def authenticate(self, request):
             # we can get from the www-authentication header) in the response.
             if resp.status_code == 401 and 'www-authenticate' in resp.headers:
                 data = parse_www_authenticate_header(resp.headers['www-authenticate'])
-                raise exceptions.AuthenticationFailed(data['error_description'])
+                raise exceptions.AuthenticationFailed(data.get('error_description'))
```
peterbe, line 82: Could that be:
```suggestion
                raise exceptions.AuthenticationFailed(data.get('error_description') or 'no error_description in www-authenticate')
```
just so that you at least can't end up with a `raise exceptions.AuthenticationFailed(None)`

HISTORY.rst
```diff
@@ -8,7 +8,8 @@ pending
 
 * Make `get_or_create_user` compatible with custom scope configuration
   by moving scope specific code to `describe_user_by_claims`
-
+* Fix bug in contrib.drf.OIDCAuthentication where `error_description` is assumed to
+  be present in `www-authenticate` header despite it being optional in the spec
```
peterbe, line 12:
```suggestion
  be present in ``www-authenticate`` header despite it being optional in the spec.
  
```

mozilla_django_oidc/contrib/drf.py
akatsoulas: or something like Peter's suggestion with `get`

```python
data.get('error_discription', 'no error description in www-authenticate')
```

akatsoulas (commented):
This looks good to me. With the changes that @peterbe suggested this should be good to go

akatsoulas:
Updated in #471

PR #432, opened 2021-08-18
add setting which adds the response from the token endpoint to a user's claims

LeoMcA:
useful/necessary for storing a refresh_token in a User model

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -286,7 +286,14 @@ class OIDCAuthenticationBackend(ModelBackend):
         if payload:
             self.store_tokens(access_token, id_token)
             try:
-                return self.get_or_create_user(access_token, id_token, payload)
+                get_or_create_kwargs = {}
+                if self.get_settings('OIDC_ADD_TOKEN_INFO_TO_USER_CLAIMS', False):
+                    # avoid modifying the method signature if the setting isn't enabled
+                    # to avoid breaking subclasses which haven't included any kwargs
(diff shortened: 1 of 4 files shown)


mozilla_django_oidc/auth.py
```diff
@@ -286,7 +286,14 @@ def authenticate(self, request, **kwargs):
         if payload:
             self.store_tokens(access_token, id_token)
             try:
-                return self.get_or_create_user(access_token, id_token, payload)
+                get_or_create_kwargs = {}
+                if self.get_settings('OIDC_ADD_TOKEN_INFO_TO_USER_CLAIMS', False):
```
akatsoulas, line 290: My first reaction here is that we are mixing different things. I totally understand the value of retrieving the `refresh_token` but this should not be mixed with user related endpoints/activities. 

Another possible direction here would be to add a hook specifically for the refresh token similar to the `update_user` method. If a setting is present then it could a method like `store_refresh_token` which will do nothing for now. Where and how the token will be stored will be up to the user.

akatsoulas: Another possible way could be to pass the refresh_token in the `store_tokens` method and do nothing with it. The user will be responsible for manipulating that.

akatsoulas:
After chatting a bit with @LeoMcA on this PR, `refresh_token` can be accessed through the `get_token` method which seems to be sufficient. We can revisit an approach specifically designed for the refresh_token in the future if there is a need for it that it's not covered by the current implementation. Thanks @LeoMcA for this PR!!

PR #436, opened 2021-08-31
Update settings.rst

bgroff:
If this is older information, it might be good to remove it. Skimmers (read me) might only see the last sentence and not read carefully enough to not use oidc_id_token as the session key on the first go round.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -172,7 +172,6 @@ of ``mozilla-django-oidc``.
    :default: ``False``
 
    Controls whether the OpenID Connect client stores the OIDC ``id_token`` in the user session.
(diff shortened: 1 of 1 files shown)


akatsoulas:
@bgroff thanks for this PR. The information in the documentation is accurate and I believe it's the only place that mentions which is the session key for the `id_token`. Probably removing it will add more confusion instead of helping people. Because of this I am closing this PR. Please do re-open if I misunderstood the purpose of the change

bgroff:
Sorry for the noise. I read that the key is now called id_token but used to be called oidc_id_token. That is not the case, I just read it wrong. Sorry again.

PR #448, opened 2022-01-17
fix: add missing extra params to SessionRefresh

melanger:
otherwise refreshing authentication is missing the extra params that the first authentication has (in the View)


akatsoulas (approved):
thanks @melanger !

Issue #450, opened 2022-02-05
Add idp_hint to /oidc/authenticate?


markperri:
Is it possible to add an extra query (idp_hint) when opening /oidc/authenticate?  I would like to change the query depending on the user logging in.  I saw #439 but it didn't seem like that was merged or customizable.

sklarsa:
I'm looking into this as well, using a different querystring param, but same idea.  I don't think it's possible to make this dynamic on a per-request basis, but I'm making a PR that adds this functionality since it seems to be a pretty common thing among IDPs

jellestoel:
Maybe this helps: I use kc_idp_hint in the following way.

settings.py. 
```
OIDC_AUTHENTICATE_CLASS = "base.views.CustomOIDCAuthenticationRequestView"
```


base/views.py
```
class CustomOIDCAuthenticationRequestView(OIDCAuthenticationRequestView):
    def get_extra_params(self, request):
        extra_params = self.get_settings("OIDC_AUTH_REQUEST_EXTRA_PARAMS", {})
        ip = request.META.get("HTTP_X_REAL_IP", None) # proxy_set_header X-Real-IP $remote_addr; in nginx
        if request.get_host() == "myapp.intra.mydomain.tld" or (
            ip and ip in settings.COMPANY_IPS
        ):
            extra_params["kc_idp_hint"] = "keycloak-oidc"
        return extra_params
        
```

markperri:
Thanks, I'll give that a try!

PR #451, opened 2022-03-14
Adds variable extra_params from callback

sklarsa:
fixes: https://github.com/mozilla/mozilla-django-oidc/issues/450

Allows implementers to customize extra_params in the authorization request based attributes from the `request` object.  This is useful in cases where IDPs expect additional querystring params in their `/authorize` APIs that are user-specific, instead of adding the same `OIDC_AUTH_REQUEST_EXTRA_PARAMS` on every request.

I added a test and updated the docs, please let me know what else I need to do!

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -180,6 +180,24 @@ of ``mozilla-django-oidc``.
 
    Additional parameters to include in the initial authorization request.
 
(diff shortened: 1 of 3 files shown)


sklarsa:
This actually isn't going to work for what I need.  I think I need to be able to pass additional querystring params to the `/oidc/authenticate/` endpoint that get passed through to the `get_extra_params` method, which this doesn't do.

PR #452, opened 2022-03-15
LOGIN_REDIRECT_URL should accept an URL or named URL pattern

dispiste:
LOGIN_REDIRECT_URL is a general django setting. According to Django docs, it should accept an URL or named URL pattern (see https://docs.djangoproject.com/en/stable/ref/settings/#login-redirect-url). However, if we use a named URL pattern, mozilla-django-oidc does not work properly.

The solution uses django.shortcuts resolve_url method, which is not publish API as far as I know. However, it is available at least in Django 2.2 to 4.x.


akatsoulas (approved):
Thanks @dispiste !

PR #469, opened 2022-11-03
Update installation.rst to set LOGIN_URL

ajcollins:
Update docs to set LOGIN_URL, which enables django's login_required decorator to use mozilla-django-keycloak's login URL.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -85,6 +85,10 @@ Start by making the following changes to your ``settings.py`` file.
    AUTHENTICATION_BACKENDS = (
        'mozilla_django_oidc.auth.OIDCAuthenticationBackend',
        # ...
+       
+   # Configure django's login_required decorator
+   # to use mozilla-django-keycloak's login URL
+   LOGIN_URL = 'oidc_authentication_init'
    )
 
 You also need to configure some OpenID Connect related settings too.


akatsoulas:
@ajcollins thank you for this PR!

Could you maybe explain why this needs to be part of the documentation? Probably there is a typo here too as you mention `keycloak` in your commit

ajcollins:
@akatsoulas thank you for considering it and for the very helpful project. Sorry for the lack of clarity and typo.

I proposed the change as, after following the mozilla_django_oidc installation instructions, views decorated with django's [`@login_required`](https://docs.djangoproject.com/en/4.1/topics/auth/default/#the-login-required-decorator
) decorator did not redirect me to authenticate through the OIDC provider. Instead redirecting to the default `/accounts/login/`.

I realise that many will add OIDC support having already set [`LOGIN_URL`](https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-LOGIN_URL) and realise that they need to update this. However, it wasn't obvious when adding OIDC support at the start of the project. Including it in the installation instructions would have set me on happy-path straight away.

I've fixed the typo and updated the comment in the code. Hopefully this improves the clarity.

akatsoulas:
Thanks @ajcollins for clarifying this!

docs/installation.rst
```diff
@@ -85,6 +85,12 @@ Start by making the following changes to your ``settings.py`` file.
    AUTHENTICATION_BACKENDS = (
        'mozilla_django_oidc.auth.OIDCAuthenticationBackend',
        # ...
+
+   # Set LOGIN_URL to mozilla_django_oidc's login URL.
```
akatsoulas, line 89: @ajcollins can you maybe rephrase this a bit to indicate that is more of a hint and not a necessary step?

ajcollins: @akatsoulas sorry for the delay. Good point. PR updated.

PR #474, opened 2022-11-16
Support for PKCE

akatsoulas:
Adjust comments

Remove .devcontainer

Default challenge method to S256

Fix base64 encoding

PKCE Unit Tests

Fix code_verifier KeyError

Test code_verifier parameter

AuthoricationRequestView test code_verifier

Replace assert_ with assertTrue

Replace all assert_ calls with assertTrue

Fix flake8 errors

Fix flake8 warnings

Test PKCE settings

Check for AssertionError rather than ValueError

Update documentation

Change AssertionError to ValueError

### DIFF
--- a/mozilla_django_oidc/utils.py
+++ b/mozilla_django_oidc/utils.py
@@ -1,12 +1,13 @@
 import logging
 import time
 import warnings
-
-from django.conf import settings
-from django.core.exceptions import ImproperlyConfigured
-
+from hashlib import sha256
 from urllib.request import parse_http_list, parse_keqv_list
 
+# Make it obvious that these aren't the usual base64 functions
+import josepy.b64
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


akatsoulas:
@escattone it would be nice to have a second pair of eyes in this one

mozilla_django_oidc/utils.py
```diff
@@ -54,16 +55,75 @@ def is_authenticated(user):
...
+
+    elif method == "S256":
+        return base64_url_encode(sha256(code_verifier.encode("ascii")).digest())
+
+    else:
+        raise ValueError("code challenge method must be 'plain' or 'S256'.")
+
+
+def add_state_and_verifier_and_nonce_to_session(
+    request, state, params, code_verifier=None
```
escattone, line 109: Nit. Why not look for `code_verifier` in the `params`, and use `code_verifier = params.get("code_verifier")` to get its value?

escattone (changes requested):
A minor suggested change in `test_utils.py`, and a question, but otherwise it looks good to me.

escattone (dismissed):

mozilla_django_oidc/utils.py
akatsoulas: Valid point but let's optimize one this is out. It's already waiting for a while now and I would prefer to get this out sooner than later

escattone (approved):
r+

PR #476, opened 2022-11-16
Add pre-commit to the project

akatsoulas:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,22 @@
+default_language_version:
+  python: python3.7
```
escattone, line 2: Since this will be running on the user's machine, it seems better to keep this as general as possible, or am I missing something? For example, specifying `python3.7` caused `pre-commit` to fail on my machine because I only have Python 3.10 installed.
```suggestion
default_language_version:
  python: python3
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+      - id: check-merge-conflict
+      - id: debug-statements
+      - id: trailing-whitespace
+      - id: check-ast
+  - repo: https://github.com/psf/black
+    rev: 22.10.0
+    hooks:
+      - id: black
+        files: "mozilla-django-oidc/|integration_tests/|tests/"
+        language_version: python3.7
```
escattone, line 18:
```suggestion
        language_version: python3
```

escattone (changes requested):
In addition to the suggestions below, I think we should also include in this PR the changes needed to ensure a clean pre-commit:
```sh
(mozilla-django-oidc) repos/mozilla-django-oidc % pre-commit run --all-files
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/pycqa/flake8.git.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
check for added large files..............................................Passed
check for case conflicts.................................................Passed
check for merge conflicts................................................Passed
debug statements (python)................................................Passed
trim trailing whitespace.................................................Failed
- hook id: trailing-whitespace
- exit code: 1
- files were modified by this hook

Fixing docs/installation.rst
Fixing HISTORY.rst
Fixing CODE_OF_CONDUCT.md

check python ast.........................................................Passed
black....................................................................Passed
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

docs/conf.py:14:11: E401 multiple imports on one line
docs/conf.py:27:1: E402 module level import not at top of file
docs/conf.py:28:1: E402 module level import not at top of file
docs/conf.py:189:5: E265 block comment should start with '# '
docs/conf.py:191:5: E265 block comment should start with '# '
docs/conf.py:193:5: E265 block comment should start with '# '
```

.pre-commit-config.yaml
akatsoulas: We need to ensure compatibility with Python 3.7. If a user submits a PR with valid Python 3.10 there is a chance that it won't be compatible with Python 3.7. For example a `match` statement

akatsoulas: We need to ensure compatibility with Python 3.7. If a user submits a PR with valid Python 3.10 there is a chance that it won't be compatible with Python 3.7. For example a `match` statement

PR #490, opened 2023-05-08
Use unittest.mock rather than mock

traylenator:
Since python 3.4 the backported module mock is no longer needed and the builtin unittest.mock can be used instead.


akatsoulas (approved):
Thanks @traylenator!

PR #496, opened 2023-06-22
move calls to `get_settings` outside of `__init__`

### LINKED ISSUE #495: Don't call `get_settings` inside __init__
### Proposal
Move calls to `get_settings` from the `__init__` of the auth backend, the middleware and the authentication request view, to a `cached_property` that returns a dictionary, or something else. This way the settings also get initialised once and it shouldn't affect the rest of this package.

### Why?
Because if you want to overwrite `get_settings`, to for example use some session data (and therefore the request object), it's impossible now. The request is not known inside the `__init__` (of views, middleware and auth backends) yet.

(The usecase I ran into is supporting multiple providers, where the chosen provider is in the session)

phijma-leukeleu:
View, middleware and auth backend

Fixes #495

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -9,6 +9,7 @@ from django.contrib.auth.backends import ModelBackend
 from django.core.exceptions import ImproperlyConfigured, SuspiciousOperation
 from django.urls import reverse
 from django.utils.encoding import force_bytes, smart_bytes, smart_str
+from django.utils.functional import cached_property
 from django.utils.module_loading import import_string
 from josepy.b64 import b64decode
 from josepy.jwk import JWK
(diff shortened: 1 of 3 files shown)


mozilla_django_oidc/middleware.py
```diff
@@ -13,10 +13,8 @@
 
 from mozilla_django_oidc.auth import OIDCAuthenticationBackend
 from mozilla_django_oidc.utils import (
-    absolutify,
-    add_state_and_verifier_and_nonce_to_session,
-    import_from_settings,
-)
+    absolutify, add_state_and_verifier_and_nonce_to_session,
+    import_from_settings)
```
ThomasKalverda, line 17:
```suggestion
    absolutify, 
    add_state_and_verifier_and_nonce_to_session,
    import_from_settings
)
```

mozilla_django_oidc/views.py
```diff
@@ -12,11 +12,8 @@
 from django.views.generic import View
 
 from mozilla_django_oidc.utils import (
-    absolutify,
-    add_state_and_verifier_and_nonce_to_session,
-    generate_code_challenge,
-    import_from_settings,
-)
+    absolutify, add_state_and_verifier_and_nonce_to_session,
+    generate_code_challenge, import_from_settings)
```
ThomasKalverda, line 16:
```suggestion
    absolutify,
    add_state_and_verifier_and_nonce_to_session,
    generate_code_challenge,
    import_from_settings
)
```

mozilla_django_oidc/middleware.py
```diff
@@ -13,9 +13,9 @@
 
 from mozilla_django_oidc.auth import OIDCAuthenticationBackend
 from mozilla_django_oidc.utils import (
-    absolutify,
+    absolutify,
```
phijma-leukeleu, line 16:
```suggestion
    absolutify,
```

mozilla_django_oidc/middleware.py
```diff
@@ -13,9 +13,9 @@
 
 from mozilla_django_oidc.auth import OIDCAuthenticationBackend
 from mozilla_django_oidc.utils import (
-    absolutify,
+    absolutify, 
     add_state_and_verifier_and_nonce_to_session,
-    import_from_settings,
+    import_from_settings
```
phijma-leukeleu, line 18:
```suggestion
    import_from_settings,
```

mozilla_django_oidc/views.py
```diff
@@ -15,7 +15,7 @@
     absolutify,
     add_state_and_verifier_and_nonce_to_session,
     generate_code_challenge,
-    import_from_settings,
+    import_from_settings
```
phijma-leukeleu, line 18:
```suggestion
    import_from_settings,
```

mozilla_django_oidc/auth.py
```diff
@@ -44,21 +45,27 @@ class OIDCAuthenticationBackend(ModelBackend):
...
-        self.OIDC_RP_SIGN_ALGO = self.get_settings("OIDC_RP_SIGN_ALGO", "HS256")
-        self.OIDC_RP_IDP_SIGN_KEY = self.get_settings("OIDC_RP_IDP_SIGN_KEY", None)
-
-        if self.OIDC_RP_SIGN_ALGO.startswith("RS") and (
-            self.OIDC_RP_IDP_SIGN_KEY is None and self.OIDC_OP_JWKS_ENDPOINT is None
+        self.UserModel = get_user_model()
+
+    @cached_property
+    def settings(self):
+        retval = {
```
sergei-maertens, line 52: @stevenbal is this going to break our library?

stevenbal: With the way our library is implemented now, yes, because we rely on overriding `get_settings`

sergei-maertens: I don't think this is too much of a problem? I don't see any changes to that method in particular and instead it will allow us to remove our `__init__` override since settings are now consumed lazily. So, what am I missing? :smile:

stevenbal: Whoops, I skimmed over it too quickly and thought `get_settings` was removed here altogether :sweat_smile:. You're right, this should work and the init override can be removed.

PR #499, opened 2023-07-11
Added on-demand session token refresh

drwonky:
Added a special case that allows AJAX queries to refresh the session on demand.  AJAX queries simply need to add the
header 'X-Refresh-OIDC-Token' to the request and it will reset the session expiration so it generates a reauth redirect. This allows SPAs and other Javascript driven applications to proactively control session refresh.


drwonky:
See PR #364 for additional context.  This is me cleaning up that PR, even if it's 3 years later :wink:

PR #500, opened 2023-07-11
Do not set oidc_login_next for AJAX 403 redirects

drwonky:
Moved the oidc_login_next to below the AJAX return, this means AJAX requests will be redirected to the LOGIN_REDIRECT_URL instead of the API endpoint that generated the session refresh.


drwonky:
See PR #364 for additional context.  This is me cleaning up that PR, even if it's 3 years later :wink:

PR #511, opened 2023-12-22
disable PKCE by default

escattone:
Not all authorization providers yet support PKCE layered on top of the normal, secret-based authorization code flow -- for example, [Mozilla Accounts explicitly forbids it](https://github.com/mozilla/fxa-auth-server/blob/master/fxa-oauth-server/docs/api.md#request-parameters-7) -- so in order to avoid the next release being a potentially breaking change, let's set the default value of `OIDC_USE_PKCE` to `False`.


akatsoulas (approved):
Thanks @escattone

Issue #397, opened 2021-03-03
Code authorization flow with PKCE


rogerbn:
Could support for Code authorization flow with PKCE be added to mozilla-django-oidc ? or does that go against design priciple "Mainly support OIDC authorization code flow"`?

krzysieqq:
@johngian Do you plan to review and merge pull request?

maxb:
PKCE is now widely recommended for use in more contexts than that in more recent standards.

escattone:
@maxb Do you know of a recommended use case for the PKCE flow on the server-side? It seems to me that on the server side of things, the "Authorization Code" flow, with its uses of a private client secret instead of PKCE's random code verifier and code challenge, would always be preferred? Perhaps server-side PKCE is a nice option because it doesn't depend on a static secret that could be stolen? Could you point us to where PKCE is recommended on the server side?

maxb:
An RFC documenting OIDC 2.1, a rollup of the many documents that have tweaked bits of OIDC 2.0 since its creation, is currently in draft with the IETF: https://datatracker.ietf.org/doc/draft-ietf-oauth-v2-1/

PKCE is now recommended for **all** usage of the Authorization Code flow:

> 7.6.  Authorization Codes
> 
>    To prevent injection of authorization codes into the client, using
>    code_challenge and code_verifier is REQUIRED for clients, and
>    authorization servers MUST enforce their use, unless both of the
>    following criteria are met:
> 
>    *  The client is a confidential client.
> 
>    *  In the specific deployment and the specific request, there is
>       reasonable assurance by the authorization server that the client
>       implements the OpenID Connect nonce mechanism properly.
> 
>    **In this case, using and enforcing code_challenge and code_verifier is
>    still RECOMMENDED.**
>
>    The code_challenge or OpenID Connect nonce value MUST be transaction-
>    specific and securely bound to the client and the user agent in which
>    the transaction was started.  If a transaction leads to an error,
>    fresh values for code_challenge or nonce MUST be chosen.
> 
>    **Historic note: Although PKCE [RFC7636] (where the code_challenge and
>    code_verifier parameters were created) was originally designed as a
>    mechanism to protect native apps, this advice applies to all kinds of
>    OAuth clients, including web applications and other confidential
>    clients.**

(Bold emphasis my own.)

escattone:
@maxb I wasn't aware of that, thanks for sharing that link to the RFC doc and enlightening me!

[...]

PR #512, opened 2023-12-23
add PKCE to SessionRefresh middleware and refactor

escattone:
This PR addresses https://github.com/mozilla/mozilla-django-oidc/issues/397#issuecomment-1432588711.

### DIFF
--- a/mozilla_django_oidc/utils.py
+++ b/mozilla_django_oidc/utils.py
@@ -2,12 +2,15 @@ import logging
 import time
 import warnings
 from hashlib import sha256
+from urllib.parse import urlencode
 from urllib.request import parse_http_list, parse_keqv_list
 
 # Make it obvious that these aren't the usual base64 functions
 import josepy.b64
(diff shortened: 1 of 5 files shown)


mozilla_django_oidc/utils.py
```diff
@@ -159,3 +162,88 @@ def add_state_and_verifier_and_nonce_to_session(
         "nonce": nonce,
         "added_on": time.time(),
     }
+
+
+class AuthorizationCodeRequestMixin:
```
akatsoulas, line 167: Although it makes a lot of sense to introduce this mixin and use it in both places, I am concerned that introducing this change will also introduce complexity and it will make it harder to modify/debug/override code. Thoughts?

escattone: @akatsoulas and I discussed the pros and cons of this via video. Closing this in favor of https://github.com/mozilla/mozilla-django-oidc/pull/515.

PR #521, opened 2024-02-14
Support application/jwt in userinfo endpoint

### LINKED ISSUE #517: Authentication backend get_userinfo incorrectly assumes the endpoint responds with application/json
https://github.com/mozilla/mozilla-django-oidc/blob/bf0d1439da144daa061128a5c857af162b904cc7/mozilla_django_oidc/auth.py#L282

This line tries to decode the response content as JSON, which is one of the ways this endpoint may be implemented. The other way is that it responds with a JWT, having a `Content-Type` header of `application/jwt;charset=utf-8` and that fails to decode as JSON, as the JWT itself needs to be processed.

Environment details:

* OS: Arch Linux
* Python: 3.10
* Django: 3.2 LTS
* mozilla-django-oidc: 3.0.0

How to reproduce:

* Use an `OIDC_OP_USER_ENDPOINT` that responds with a JWT
* Authenticate using OIDC
* Observe `JSONDecodeError` crash

sergei-maertens:
Closes #517

This is an initial draft to spark discussion about implementation details.

Changes:

* Refactored the JWS verification into a generic helper - the goal is that the middleware methods are backwards compatible
* Added (partial) handling for `application/jwt` processing in the userinfo endpoint

Topics to discuss:

* test failures due to missing Content-Type header -> can we use a utility like vcr.py to record real requests instead of using unittest.mock / request mocking approaches? My experience is that those tools are very fragile
* I have a keycloak setup that can be added to the docker-compose which is configured for application/jwt - this is a powerful test setup with vcr.py from the above bullet
* How do we handle keys that are not loaded from a JWKS endpoint? I think a new setting would be relevant, but is that even a case you want to support and/or is used in real deployments? Looking at the code again, I'm not sure that HMAC 256 is even configurable?
* I've used type annotations because they help me write better code - some maintainers have strong opinions about these, just let me know your position about this and I'll adapt.
* Is this the right direction? It works like a breeze for us in our downstream library, but I'm open to feedback of course
* How do you feel about block-listing algorithm `none`?
* I looked into re-using `parse_www_authenticate_header` for the content type header processing, but it kinda borks on a value like `application/json; charset=utf-8`, so instead I use the (private) utility from the `requests` library which can be controversial. I don't trust myself enough to correctly and safely parse HTTP headers :grimacing:

### DIFF
--- a/mozilla_django_oidc/_jose.py
+++ b/mozilla_django_oidc/_jose.py
@@ -1,0 +1,79 @@
+"""
+Helpers for dealing with Javascript Object Signing and Encryption (JOSE) in an OpenID
+Connect context.
+"""
+import json
+from typing import Dict, Any, TypeAlias, Union
+
+from django.core.exceptions import SuspiciousOperation
+from django.utils.encoding import smart_bytes
+
+from josepy.jwk import JWK
+from josepy.jws import JWS
+
+# values could be narrowed down to relevant JSON types.
+Payload: TypeAlias = Dict[str, Any]
+
+
+def verify_jws_and_decode(
+    token: bytes,
+    key,
+    signing_algorithm: str = "",
+    decode_json: bool = False,  # for backwards compatibility reasons
+) -> Union[Payload, bytes]:
+    """
+    Cryptographically verify the passed in token and return the decoded payload.
+
+    Verification is done with utilities from the josepy library.
+
+    :arg token: the raw binary content of the JWT/token.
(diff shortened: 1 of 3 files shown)


ikarius:
Seems legit for the handling of the `application/jwt' content-type. 
We stumbled on this issue where JWT content was in binary/byte array form, and the `userinfo` response was unable to be decoded (because not in standard JSON).

sergei-maertens:
Fantastic news! 🎉

PR #526, opened 2024-02-29
Add support for Django 5.0

akatsoulas:


tonial (approved):

willkg (approved):
This probably looks fine.

We might want to think about switching from CircleCI to GitHub actions like most of our other projects. And I wonder if there's a better way to specify all the jobs that isn't so typing-intensive.

mmc1718:
Is this ready to be merged? I would like to use this package for an application using the latest version of Django. Without this PR merged would I have to downgrade the version of Django I am using? Thanks!

janbrasna:
I can look into moving the CI to GitHub Actions (and upgrading the versions along the way) if there's interest — in case any maintainer who would be able to review then (Tasos?) is around, just let me know and I'll start digging into it.

_(Esp. if there's a major version release planned, I'd chime in and drop EOL versions in text matrix, and add some newer ones while at it…)_

@mmc1718 You don't have to downgrade, we're running it on the latest 5.2.x versions successfully — it's just the visibility here into potential change impact is not added for newer Django releases.

Issue #539, opened 2024-10-15
OpenSSL Deprecation warning


tonial:
OpenSSL recently added a depreciation warning on some internal methods.

```
inclusion_connect/oidc_federation/base.py:13: in <module>
    from mozilla_django_oidc import auth, views
../.virtualenv/inclusion-connect/lib/python3.11/site-packages/mozilla_django_oidc/auth.py:14: in <module>
    from josepy.b64 import b64decode
../.virtualenv/inclusion-connect/lib/python3.11/site-packages/josepy/__init__.py:40: in <module>
    from josepy.json_util import (
../.virtualenv/inclusion-connect/lib/python3.11/site-packages/josepy/json_util.py:26: in <module>
    from josepy import b64, errors, interfaces, util
../.virtualenv/inclusion-connect/lib/python3.11/site-packages/josepy/util.py:18: in <module>
    class ComparableX509:
../.virtualenv/inclusion-connect/lib/python3.11/site-packages/josepy/util.py:26: in ComparableX509
    def __init__(self, wrapped: Union[crypto.X509, crypto.X509Req]) -> None:
../.virtualenv/inclusion-connect/lib/python3.11/site-packages/cryptography/utils.py:70: in __getattr__
    warnings.warn(obj.message, obj.warning_class, stacklevel=2)
E   DeprecationWarning: CSR support in pyOpenSSL is deprecated. You should use the APIs in cryptography.
================================= short test summary info =================================
ERROR tests/oidc_federation/test_peama.py - DeprecationWarning: CSR support in pyOpenSSL is deprecated. You should use the APIs in cryptography.
```

Josepy maintainers are thinking about dropping the library (see https://github.com/certbot/josepy/pull/182#issuecomment-2356757648 and https://github.com/certbot/certbot/issues/8322#issue-709307708) 

Since https://github.com/jpadilla/pyjwt seems to have more commits/releases recently I'll try to replace josepy with it in a PR.

EDIT -> done in #543

stevejalim:
@akatsoulas Will there be a new release coming soon, please?

PR #543, opened 2024-10-19
Replace josepy with PyJWT

tonial:
Also fixes https://github.com/mozilla/mozilla-django-oidc/issues/539 but with PyJWT instead of jwcrypto as it seems to be better maintained

I put the changes in the tests in a separated commit to ensure the changes in auth.py didn't break any test prior to changing them.


akatsoulas (approved):
Thanks @tonial for this PR! Much appreciated

Issue #548, opened 2025-03-31
Allow parsing for JWT userinfo response


ikarius:
## Issue 

As mentioned by OIDC specs, return value of the `UserInfo` endpoint can be a JWT.

Current implementation of `OIDCAuthenticationBackend.get_userinfo` is only able to parse a JSON result.

## Proposal

Check if `UserInfo` return type is `application/jwt` and decode accordingly.

escattone:
This is released.

PR #549, opened 2025-03-31
Ability to parse JWT UserInfo payload

ikarius:
A possible fix for #548 + simple content-type test

### DIFF
--- a/mozilla_django_oidc/auth.py
+++ b/mozilla_django_oidc/auth.py
@@ -279,6 +279,12 @@ class OIDCAuthenticationBackend(ModelBackend):
             proxies=self.get_settings("OIDC_PROXY", None),
         )
         user_response.raise_for_status()
+
+        if user_response.headers.get("content-type", "").startswith("application/jwt"):
+            # OIDC userinfo claims can be encoded as JWT
+            return self.verify_token(user_response.text)
+
+        # otherwise process as JSON payload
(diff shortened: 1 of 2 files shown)


ikarius:
Mmm, e2e testing is failing, but it does not seem to be code related :
```
    raise DriverNotFoundError(f"Driver for {driver_name} was not found.")
splinter.exceptions.DriverNotFoundError: Driver for firefox was not found.
``

File: mozilla_django_oidc/auth.py [hunk 1]
```diff
@@ -279,6 +279,12 @@ def get_userinfo(self, access_token, id_token, payload):
             proxies=self.get_settings("OIDC_PROXY", None),
         )
         user_response.raise_for_status()
+
+        if user_response.headers.get("content-type", "").startswith("application/jwt"):
```
qbey, line 283: MIME types are case-insensitive you may want to be also?

> MIME types are case-insensitive but are traditionally written in lowercase. The parameter values can be case-sensitive.
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types

```suggestion
        if user_response.headers.get("content-type", "").lower().startswith("application/jwt"):
```

qbey (commented):
:pray:

escattone (approved):

PR #551, opened 2025-07-25
auth: Skip OIDC flow if Bearer access_token is present (bug 1979246)

shtrom:


shtrom:
E2E tests seem to fail due to a firefox (web?)driver missing.

janbrasna:
(Yeah the CI is now updated) Would you mind rebasing that for fresh test runs @shtrom? TY!

shtrom:
Just rebased and pushed 🤞🏼

shtrom:
> I would suggest putting this functionality in a separate backend that extends `OIDCAuthenticationBackend`. Something like `OIDCTokenAuthenticationBackend`.

Ok, I did just that in the last commit I pushed!

This way both authentication methods can be selectively composed as needed in the standard Django fashion.

shtrom:
The PR title should probably be changed to something like 

> auth: add OIDCTokenAuthenticationBackend to authenticate requests based on Bearer access_token (bug 1979246)

But I don't have permissions to do so.

PR #573, opened 2025-12-18
Update RTD config to fix docs pipeine

janbrasna:
Removed `requirements/requirements_dev.txt` fail RTD builds, uses current `.[docs]` instead.

Also bumps OS version.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -6,7 +6,7 @@ version: 2
 
 # Set the OS, Python version and other tools you might need
 build:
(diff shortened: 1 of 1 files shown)


janbrasna:
Cool, that worked! https://mozilla-django-oidc.readthedocs.io/en/latest/history.html 🙏 💪 

(too bad they don't re-generate failed `stable` builds to back-fill the missing deployments, so this currently only covers `latest` ("dev") — but good it's running now.)

Issue #579, opened 2026-02-26
Token invalidation due to server time and token 1 minute difference


lucas8107:
I am getting the iat validation error because the token iat is 1 minute in the future, maybe add a config to set the leeway time?

lucas8107:
https://github.com/mozilla/mozilla-django-oidc/blob/86660c22c842d935551e3da7ab14709e53facd12/mozilla_django_oidc/auth.py#L144-L146 we could pass the config value here to the jwt api

fabien-michel:
I got the same issue with less than a second difference:
```python
iat = datetime.datetime(2026, 5, 11, 12, 38, 24)
now = datetime.datetime(2026, 5, 11, 12, 38, 23, 746832)
```
Thanks for the PR.

Here a monkey patch waiting for the PR to be validated:

```python
from django.apps import AppConfig


class MyAppConfig(AppConfig):
    name = "my_app"

    def ready(self):
        monkeypatch_pyjwt_leeway()

def monkeypatch_pyjwt_leeway():
    import jwt

    _original_decode_complete = jwt.PyJWT.decode_complete

    def patched_decode_complete(self, *args, **kwargs):
        kwargs["leeway"] = 5
        return _original_decode_complete(self, *args, **kwargs)

    jwt.PyJWT.decode_complete = patched_decode_complete
```

PR #580, opened 2026-03-20
Add leeway config for PyJWT's jwt.decode function

yaourtv:
This PR adds `OIDC_JWT_LEEWAY` setting to pass a leeway value to the `jwt.decode(leeway=...)` function to avoid errors like 'The token is not yet valid (iat)' (#579) on clock skew.


pepijnolivier:
@akatsoulas 

Any chance we could get this one merged in ? 🙏 
Adding a bit more leeway helps when the keycloak server time is slightly off from the clock.

